PR #20, opened 2021-07-05
Add Zenodo example

peterdesmet:
@damianooldoni since the functionality is still small, I added an example to the README, rather than starting a vignette.

Note, my link to Zenodo does not link to that specific dataset, but shows all datasets that have the `frictionlessdata` keyword.


damianooldoni (approved):
Nice! Woud you consider to update the version too? I do it typically at any PR, even if small and only related to documentation.

PR #62, opened 2021-12-03
Make `get_resource()` an internal function

peterdesmet:
This PR makes `get_resource()` an **internal function**.

It is used internally by the user facing functions `read_resource()`, `remove_resource()`, `write_package()` and `get_schema()`. Although it could argued that it is useful for a user to get the whole list object for a Data Resource using this function, the **returned object cannot be used by any of the other functions**. I think it is therefore odd to list it with the rest of the functions and should be kept internal. Note that users can get any part of the Data Package object by navigating its list structure, e.g. `pkg$resources[[1]]`.

Note that the similarly named `get_schema()` will likely be useful to keep public, since `schema` will be a parameter in `add_resource(df, schema). So there is a workflow where users 1) get a schema, 2) augment it (e.g. add definitions) and 3) use that when adding a resource.

Even though the function has documentation, it is marked with `@noRd` so no man file will be created. I didn't want to start mixing `@noRd` with `@keywords internal` (which does create a man file but hides it from the index).

This PR also moves the code for internal function `check_package()` to its own file.


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: frictionless
 Title: Read and Write Frictionless Data Packages
-Version: 0.5.0.9003
+Version: 0.5.0.9004
```
damianooldoni, line 3: Just a detail: I think it should be 0.5.1 based on https://r-pkgs.org/release.html#release-version

damianooldoni (approved):
Nice work, @peterdesmet. Just consider my note about versioning: this PR is definitely more than just a patch.

PR #68, opened 2021-12-17
Remove use of %>%

peterdesmet:
Removes the (only remaining) direct import for `dplyr %>%`. It is only used in one helper function. Rather than using `dplyr::%>%` in that function, I opted to nest the function calls. Slightly less readable, but not very complex.


damianooldoni:
Yes, if it's used just in one place, it doesn't make sense to have it.

damianooldoni (approved):
Correctly done!

PR #69, opened 2021-12-17
Replicate rlist::list.clean as helper function list_clean

### LINKED ISSUE #45: Can we remove rlist dependency?
We currently use `rlist` for one step:

https://github.com/frictionlessdata/frictionless-r/blob/750eef30fa1e29d5bd5611e99ae1ef748c90de95/R/create_schema.R#L117-L122

It might be worth investigating if this functionality cannot be replace with base R and purrr:: code.

peterdesmet:
Fixes #45

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: frictionless
 Title: Read and Write Frictionless Data Packages
-Version: 0.7.0.9000
+Version: 0.8.0.9000
 Authors@R: c(
     person("Peter", "Desmet", , "peter.desmet@inbo.be", role = c("aut", "cre"),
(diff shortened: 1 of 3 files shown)


damianooldoni (approved):
As the code used in such function is basic R and doesn't rely on other fcuntions of the same package or other packages, copying the function from another MIT license package works fine!
Good idea. Only thing I would change is adding a link to the rlist::list.clean() function documentation:  https://www.rdocumentation.org/packages/rlist/versions/0.4.6.1/topics/list.clean

peterdesmet:
Will use default `[rlist::list.clean()]` syntax to link. In this case it won't generate documentation because of `noRd`, but it is the default way to link to functions.

PR #76, opened 2022-01-07
Attach `csv` file with `add_resource()` + download local paths with `write_package()`

### LINKED ISSUE #74: Allow to add resources with a path
One might want to create a data package from csv files on disk. In that case it should be possible to use `write_package()` to just write the `datapackage.json` to disk and not read/write all the csv files.

On disk, in directory `my_dir`:

```
gps-2018.csv
gps-2019.csv
```

```R
my_package <- create_package()

# Create resource
gps <- read_csv("my_dir/gps-2018.csv")
add_resource(
  my_package,
  resource_name = "ref",
  df = ref, # Not used to write data, only to create schema
  path = c("my_dir/gps-2018.csv", "my_dir/gps-2019.csv") # Define path
)
# Sets resource "path", does not add inline "data"

# Write package
write_package(package, directory = "my_dir")
# Detects that some of the paths contain "my_dir" and:
# - Does not write the files to disk
# - Shortens the path to just the file names
```

Caveats:
- CSV dialect, delimiters, etc. not known.

peterdesmet:
## Attach csv file with add_resource(), fixes #74

- Allows a resource to be **added from file path(s)** rather than a data frame. The same parameter `data` is used to attach data frames or file path(s).
- Function will internally use the last provided file to make a `df`. This `df` is necessary to create or compare with a `schema`.
  - Only one file is read to speed up the process (otherwise why add csv's directly?).
  - The last file is used, since for growing datasets that is likely going to be the newest (and not yet checked) one.
  - The default `read_delim(delim = ",", progress = FALSE, show_col_types = FALSE)` behaviour can be overwritten by the user by providing more arguments in `...`
- Provided paths can be absolute or relative, since they are set by the user (and not unsafe). A new `attr` is added to the resource so that `get_resource()` does check for unsafe paths for these resources.
- The new use of the function is shown in the example and vignette.

## Also includes

- Renames `df` parameter to `data` in `create_schema()` and internal `check_schema()` as well.
- No longer refer to `df` in error messages, use `data` or data.
- Rename internal `list_clean()` to `clean_list()` (verb first).
- `observations` resource now has URL paths to serve as an example for that

~~No additional test have been added, so coverage is likely to drop. Will be added before PR can be accepted.~~ Tests are now added


R/add_resource.R
```diff
@@ -9,9 +9,17 @@
 #' `.`, `-` and `_`.
 #'
 #' @inheritParams read_resource
-#' @param df A data frame.
-#' @param schema List object describing a Table Schema for that data frame.
-#'   If not provided, one will be created (using [create_schema()]).
+#' @param data Data to attach, either a data frame or path(s) to CSV file(s).
+#'   - Data frame: attached to the resource as `data` and written to a CSV file
+#'     when using [write_package()].
+#'   - Path(s) to CSV file(s): added to the resource as `path`.
```
damianooldoni, line 15: I would add the fact that paths should be passed as a vector. Or do you want to support lists as well? I think a vector with paths is sufficient. But, it should be mentioned in documentation.

R/add_resource.R
```diff
@@ -32,21 +40,25 @@
 #'   y = c(860, 900)
 #' )
 #'
-#' # Add the data frame as a new resource to the Data Package
-#' package <- add_resource(package, "positions", df)
+#' # Add a new resource "positions" to the Data Package, from the data frame
+#' package <- add_resource(package, "positions", data = df)
 #'
-#' # List the resource names ("positions" added)
+#' # Add a new resource "deployments2" to the Data Package, from a CSV file path
```
damianooldoni, line 46: I think it would be worth to add the following examples:
1. multiple paths
2. user passes a schema via arg `schema`

R/add_resource.R
```diff
@@ -59,6 +71,28 @@ add_resource <- function(package, resource_name, df, schema = NULL) {
...
+    # Check existence of files (no further checks)
+    paths <- purrr::map_chr(
+      data, ~ check_path(.x, directory = NULL, unsafe = TRUE)
+    )
+    last_file <- paths[length(paths)]
+    # Parse arguments to be passed to read_delim and set defaults
+    read_args <- list(...)
+    read_args$file <- replace_null(read_args$file, last_file)
+    read_args$delim <- replace_null(read_args$delim, ",")
+    read_args$progress <- replace_null(read_args$progress, FALSE)
```
damianooldoni, line 91: Personally I like to show always progress bars when reading files. The reason is quite simple. 
If the file is small, you don't notice the bar even and the processing time is still very small. If the file is big, you know why you are waiting for.

damianooldoni (approved):
It seems everything is fine. I have mainly provided feedback related to the documentation. See comments.

damianooldoni (changes requested):
Yes, #77 was indeed tricky. At the end writing to disk is unavoidable.

Could you please add some message to inform the user about the progress of `write_package()`? Based on your example in #77 it seems that it can take up to minutes to download everything and you know how annoying it can be to not receive any information about the status of a function running for more than few seconds 😄 
As reference example, I have in mind the first part of the message returned by `etn::download_acoustic_dataset()`, see lines [L55-L63](https://github.com/inbo/etn/blob/main/R/download_acoustic_dataset.R#L55-L63).

peterdesmet:
@damianooldoni Download function will now display a message (4c2d1b4):

```R
> write_package(package, "zenodo")
Downloading file from https://zenodo.org/record/5070086/files/O_WESTERSCHELDE-acceleration-2019.csv
Downloading file from https://zenodo.org/record/5070086/files/O_WESTERSCHELDE-acceleration-2020.csv
```

damianooldoni (approved):
Nice work, @peterdesmet!

PR #80, opened 2022-01-13
Fix #79: create_schema() will interpret columns with all NA as string

### LINKED ISSUE #79: Set type for empty CSV fields to `string`, not `boolean` in `create_schema()`
When adding a resource from a CSV, a schema will be created for the file. `read_delim()` will interpret empty fields in the CSV files as `logical` (see https://github.com/tidyverse/readr/issues/839). For the interpretation to a Table Schema, it would be better if those were set to `string`, not `boolean`.

The only way I see on how to do that is to see if there's a way to see if all values in the column are `NA` and if so, set that column to `string`.

@damianooldoni thoughts?

peterdesmet:
This is not only done for data frames as result of `read_delim()` in `add_resource()`, but all data frames, since it is default R behaviour to set columns with all `NA` as logical.

The function `create_schema()` now handles these. Note that the data frame itself is not altered, only the field type in the resulting `schema` will be set to `string` rather than `boolean` for these columns.

I used `lapply()` with anonymous function. Alternative would be dplyr functions:

```R
dplyr::mutate(df, dplyr::across(tidyselect::where(all(is.na)), as.character))
```

but that requires tidyselect as dependency and it will raise an error for col names that are `NA` (allowed, there is a test for that).

### DIFF
--- a/R/create_schema.R
+++ b/R/create_schema.R
@@ -79,7 +79,7 @@ create_schema <- function(data) {
   # Columns with all NA are considered logical by R (and read_delim)
   # Set those to character, since string is a better default for Table Schema
   data_as_list <- lapply(data, function(x) {
-    if (all(is.na(x))) {
+    if (is.logical(x) & all(is.na(x))) {
       as.character(x)
     } else {
       x
--- a/R/create_schema.R
+++ b/R/create_schema.R
@@ -76,6 +76,16 @@ create_schema <- function(data) {
     )
   )
 
+  # Columns with all NA are considered logical by R (and read_delim)
+  # Set those to character, since string is a better default for Table Schema
+  data_as_list <- lapply(data, function(x) {
+    if (all(is.na(x))) {
+      as.character(x)
+    } else {
+      x
+    }
+  })
+
   # Create fields (a list of lists)
   fields <- purrr::imap(data_as_list, function(x, name) {
     # Name
--- a/R/create_schema.R
+++ b/R/create_schema.R
@@ -77,38 +77,37 @@ create_schema <- function(data) {
   )
 
   # Create fields (a list of lists)
-  fields <-
-    purrr::imap(data, function(x, name) {
-      # Name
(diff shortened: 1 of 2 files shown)


damianooldoni (changes requested):
See example below:

```
library(frictionless)
# Create a data frame
df <- data.frame(
  id = c(as.integer(1), as.integer(2)),
  timestamp = c(
    as.POSIXct("2020-03-01 12:00:00", tz = "EET"),
    as.POSIXct("2020-03-01 18:45:00", tz = "EET")
  ),
  life_stage = factor(c("adult", "adult"), levels = c("adult", "juvenile")),
  age = c(NA_integer_, NA_integer_)
)
# Create a Table Schema from the data frame
schema <- create_schema(df)
str(schema)
#> List of 1
#>  $ fields:List of 4
#>   ..$ :List of 2
#>   .. ..$ name: chr "id"
#>   .. ..$ type: chr "integer"
#>   ..$ :List of 2
#>   .. ..$ name: chr "timestamp"
#>   .. ..$ type: chr "datetime"
#>   ..$ :List of 3
#>   .. ..$ name       : chr "life_stage"
#>   .. ..$ type       : chr "string"
#>   .. ..$ constraints:List of 1
#>   .. .. ..$ enum: chr [1:2] "adult" "juvenile"
#>   ..$ :List of 2
#>   .. ..$ name: chr "age"
#>   .. ..$ type: chr "string"
```
Created on 2022-01-13 by the reprex package (v2.0.1)

If the data.frame contains a column full of `NA_integer_`, it should not be set to `"string"`, but to `"integer"`. A column with `NA_real_` should be set to `number`. About complex numbers I don't know if the type `"number"` encompasses it and so the corresponding `NA_complex_`.

@peterdesmet: what do you think?

peterdesmet:
Fixed in df6ce93.

damianooldoni (approved):
Nice! Approved 👍

PR #84, opened 2022-01-17
Fix #83

### LINKED ISSUE #83: Can we remove dplyr dependency
dplyr is the heaviest dependency, both in size as in number of its own dependencies. It is used 7 times in package. Might be useful to investigate if it can be removed.

``` r
library(itdepends)
library(dplyr)
library(frictionless)
uses <- suppressWarnings(itdepends::dep_usage_pkg("frictionless"))
uses %>% filter(pkg == "dplyr")
#> # A tibble: 7 × 2
#>   pkg   fun      
#>   <chr> <chr>    
#> 1 dplyr pull     
#> 2 dplyr arrange  
#> 3 dplyr desc     
#> 4 dplyr as_tibble
#> 5 dplyr as_tibble
#> 6 dplyr bind_rows
#> 7 dplyr recode
```

<sup>Created on 2022-01-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

peterdesmet:
Decided to **not drop** dplyr:

1. It's `bind_cols()` is much faster than the default `rbind()` ([source](https://stackoverflow.com/a/59482527/2463806))
2. It would require loading `tibble`, which is about the same size. `as_tibble()` and `tibble()` are exported from `dplyr` (as is `%>%`)

What was changed in this PR is an update of the util function `unique_sorted()` which now no longer requires dplyr (but just base R) and is a bit easier to read.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -18,18 +18,16 @@ replace_null <- function(x, replacement) {
 #' Get unique vector values sorted by how often they occur
 #'
 #' @param x Vector, e.g. `c("a", "b", "b", "b", "a")`.
-#' @return Vector with unique values sorted by occurrence, e.g. `c("b", "a")`.
+#' @return Vector with unique values sorted by desc. count, e.g. `c("b", "a")`.
 #' @family helper functions
 #' @noRd
 unique_sorted <- function(x) {
-  dplyr::pull(
-    dplyr::arrange(
-      # Create data.frame with values ("Group.1") and how often they occur ("x")
-      stats::aggregate(x, by = list(x), FUN = length),
-      dplyr::desc(x)
-    ),
-    "Group.1"
-  )
+  # Create data frame with values and how often they occur
+  df <- stats::aggregate(x, by = list(x), FUN = length)
+  colnames(df) <- c("value", "count")
+  # Sort dataframe on count (can only be ascending) and retrieve values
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -18,18 +18,16 @@ replace_null <- function(x, replacement) {
...
-      dplyr::desc(x)
-    ),
-    "Group.1"
-  )
+  # Create data frame with values and how often they occur
+  df <- stats::aggregate(x, by = list(x), FUN = length)
+  colnames(df) <- c("value", "count")
+  # Sort dataframe on count (can only be ascending) and retrieve values
+  values <- df[with(df, order(count)),][[1]]
+  rev(values) # Reverse order
```
damianooldoni, line 30: Maybe can you use arg `decreasing` in L29 and remove 30? To be tested...

damianooldoni (approved):
Nice. Yes, I agree with maintaining `dplyr` as package, so being able to use `as_tibble()` and `%>%` for example and so other packages as `mgrittr` and `tibble` are not needed. And if `bind_rows` is faster van basic R function `rbind`, then no doubts anymore! Just check my only comment about the use of `order` and `rev`. For the rest, keep the nice work! 💪

R/utils.R
peterdesmet: I have tried, that didn't work.

PR #85, opened 2022-01-17
Set `format`, `mediatype` and `encoding` for added CSV resources

### LINKED ISSUE #78: Add `format`, `mediatype` and `encoding` properties to resource created from CSV file
Set following properties when adding CSV file with `add_resource()` (as they might deviate from the default):

- [ ] `format`
- [ ] `mediatype`
- [ ] `encoding`

peterdesmet:
Fix #78


damianooldoni (approved):
Great work and fast!
Now that you have `utils` officially among dependencies you can use all its powerful tools which are used behind the screen by many other packages. I saw `utils::tail()` for example.

PR #87, opened 2022-01-18
Warn on unknown encodings and read with UTF-8

### LINKED ISSUE #86: `read_resource` can't read resources with encoding UTF-8 with BOM
As noted in https://github.com/frictionlessdata/specs/issues/613, the [Data Resource](https://specs.frictionlessdata.io/data-resource/#optional-properties) spec doesn't specify how the [byte order mark (BOM)](https://en.wikipedia.org/wiki/Byte_order_mark) should be handled. It is also worth noting that, probably because UTF-8 with BOM is not an encoding _per se_, it is not [registered with IANA](https://www.iana.org/assignments/character-sets/character-sets.xhtml) and therefore doesn't have a preferred MIME name.

When using `frictionless.py` describe, it will use the identifier `utf-8-sig`. However, `readr` [doesn't expect anything explicit](https://github.com/tidyverse/readr/issues/1187#issuecomment-811847866) for reading csv files with BOM, and we get the following error

``` r
library(frictionless)

dp <- read_package("https://raw.githubusercontent.com/dados-mg/datapackage-reprex/utf-8-bom/datapackage.json")

dp |> read_resource("utf-8-bom")
#> Error: Unknown encoding utf-8-sig
```

It might be a good idea to simply replace `utf-8-sig` and `utf-8-bom` (and the uppercase versions) with `utf-8`.

Ps. Although the use of [BOM is neither required nor recommended](https://en.wikipedia.org/wiki/Byte_order_mark#cite_note-4), we are currently using BOM in the files that we produce because of the large number of excel users among the consumers of our data.

peterdesmet:
Fix #86.


tests/testthat/test-read_resource.R
```diff
@@ -350,12 +350,74 @@ test_that("read_resource() understands encoding", {
...
+    )
+  )
+
+  resource <- suppressWarnings(read_resource(p, "mark_decimal"))
+  expect_identical(resource$num, expected_value) # 3000000.30
+  expect_identical(resource$num_undefined, expected_value) # 3000000.30
+
+  # Non-default decimalChar/groupChar
+  warnings <- capture_warnings(read_resource(p, "mark_decimal_group"))
+  expect_true(length(warnings) == 3) # 2 warnings + 1 parsing failure last field
```
damianooldoni, line 399: From my personal experience, this kind of checks can be fragile against refactoring.
If you can find test examples where only one warning at a time is returned instead of 2 warning and 1 parsing failure, would make the unit-tests very robust.

damianooldoni (approved):
Nice work. Just a note about unit-tests. Still, PR approved!

tests/testthat/test-read_resource.R
peterdesmet: I agree, but this one is specifically designed to raise 2 warnings (and a parsing error). So, I'll leave it.

PR #89, opened 2022-01-19
Drop `datapackage` attribute, making it easier to use `append()` to add attributes.

peterdesmet:
Also rephrases `list object` to `list`.


vignettes/frictionless.Rmd
```diff
@@ -162,7 +162,14 @@ my_package <-
   add_resource(resource_name = "observations", data = c(path_1, path_2))
 ```
 
-In the above steps you started a Data Package from scratch with `create_package()`, but you use the same functionality to edit an existing Data Package read with `read_package()`.
+Your Data Package now contains 2 resources, but you can add metadata properties as well (see the [Data Package documentation](https://specs.frictionlessdata.io/data-package/#metadata) for an overview). Since it is a list, you can use [purrr::prepend()] to insert properties at the desired place. Let's add `name` as first and `title` as second property:
+
+```{r}
+my_package <- purrr::prepend(my_package, c(name = "my_package"))
```
damianooldoni, line 168: Probably it's better to use basic R function `append()`. However, it is not in the source code of the package, so it's not  so important.

damianooldoni (approved):
Yes, as we discussed today the use of word `object` seems to me not ideal within a functional programming paradigm (we don't use "objects" in this package). Great and fast work. 🚀

vignettes/frictionless.Rmd
peterdesmet: Updated after all in vignette.

PR #91, opened 2022-01-21
Support adding a schema via path or URL in add_resource()

peterdesmet:
Noticed I needed this for a project and makes sense as a general feature.


damianooldoni (approved):
Flawless! 🥇

PR #92, opened 2022-01-24
Set format based on delimiter in `add_resource()`

### LINKED ISSUE #90: Use `format=csv` for extension `csv.gz`
For a path ending in `csv.gz`, `add_resource()` will write `format = gz`. It would be more correct if this was set to `format = csv, compression = gz`. See https://specs.frictionlessdata.io/patterns/#specification-3

peterdesmet:
This PR started as "ignoring compression extension" (first commit), but reading the specs again, I opted for a simpler approach where `format` can only be `csv` or `tsv` based on the delimiter:

https://github.com/frictionlessdata/frictionless-r/issues/90#issuecomment-1020234887

Fix #90


damianooldoni (approved):
Nice!

PR #93, opened 2022-01-24
Add test for not overwriting existing files

### LINKED ISSUE #88: Add test for not overwriting existing files
I'm relying on behaviour of `write_package()` to not overwrite any existing files, except `datapackage.json`. Would be good there is a test for that.

peterdesmet:
Fix #88.

Done by creating empty files, with size 0. When using `write_package()` to that directory, only `datapackage.json` should be altered, the other files should remain having size 0.


damianooldoni (approved):
Nice way to test it! Also quite compact and most important, creating files of dimension 0 is quite fast. PR is fully approved.

PR #94, opened 2022-01-25
Remove stringr dependency

### LINKED ISSUE #46: Can we remove stringr dependency?
Currently used once:

https://github.com/frictionlessdata/frictionless-r/blob/750eef30fa1e29d5bd5611e99ae1ef748c90de95/R/read_resource.R#L246-L248

peterdesmet:
Use `switch()` rather than `str_replace_all()` to assign datetime format (could also have used `dplyr::recode()`). The reason this works is because all statements were looking for _whole matching_ strings, which is basically a recode/switch. Only the last statement was a regex match, which is now done with `gsub()` in the default return of the switch.

Fix #46

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,6 @@ Imports:
     jsonlite,
     purrr,
     readr (>= 2.1.0),
-    stringr,
     utils,
     yaml
 Suggests:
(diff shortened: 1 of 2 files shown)


damianooldoni (approved):
Nice solution with base R functions.

PR #95, opened 2022-01-25
Use test_path() so paths work outside tests too

peterdesmet:
Inspired by https://blog.r-hub.io/2020/05/29/distribute-data/#data-outside-of-your-package

```R
testthat::test_path("data/df.csv")
#> "tests/testthat/data/df.csv"
```

Allows to copy/paste test code and run it outside of tests. `test_path()` is obviously not used for tests that specifically test for relative or absolute paths.


damianooldoni (approved):
This is the handiest commando I have learned in months! Nice catch @peterdesmet!

Issue #115, opened 2022-11-07
Skip tests on CRAN that require internet connection


peterdesmet:
Tests on CRAN might fail because of a URL request timeout (especially to Zenodo). These tests should be skipped on CRAN with `skip_if_offline()` (also skips CRAN) or `skip_on_cran()`. Note that the `example_package` requires a remote connection (see #114).

peterdesmet:
Resolved with #116

Issue #114, opened 2022-11-07
Do not require internet connection for `example_package`


peterdesmet:
`example_package` currently has a remote `$directory`:

```R
> library(frictionless)
> example_package$directory
[1] "https://raw.githubusercontent.com/frictionlessdata/frictionless-r/main/inst/extdata"
```

That means an internet connection is need to read resources. This was by design for some tests, but should probably be altered so **more tests can run without an internet connection**.

https://github.com/frictionlessdata/frictionless-r/blob/b6d330858f9c3ce1bc95e1c23520766f0ffea3d5/R/data.R#L16-L21

Need to investigate first what tests would fail (on GitHub CI) if the directory is local.

peterdesmet:
I think the better solution is to create a function `example_package()`.

This allows to clearly differentiate between remote and local (default).

```R
#' @export
example_package <- function(remote = FALSE) {
  if (remote) {
    path <- file.path(
      "https://raw.githubusercontent.com/frictionlessdata/frictionless-r",
      "main/inst/extdata/datapackage.json"
    )
  } else {
    path <- system.file("extdata", "datapackage.json", package = "frictionless")
  }
  read_package(path)
}
```

Issue #113, opened 2022-10-27
Fix CRAN error


peterdesmet:
CRAN currently shows an ERROR: https://cran.r-project.org/web/checks/check_results_frictionless.html

This is caused by a URL timeout in one of the tests (on one OS only, r-devel-linux-x86_64-fedora-gcc). It was commented we should more safely code for that.

```
Error in `open.connection(con, "rb")`: Timeout was reached: [github.com] SSL connection timeout
     Backtrace:
     ▆
     1. └─frictionless::add_resource(p, "remote", remote_path, schema) at test-add_resource.R:257:2
     2. └─readr::guess_encoding(last_file, n_max = 1000)
     3. ├─base::unlist(read_lines_raw(file, n_max = n_max))
     4. └─readr::read_lines_raw(file, n_max = n_max)
     5. └─readr::datasource(...)
     6. └─readr:::datasource_connection(...)
     7. └─readr:::read_connection(path)
     8. ├─base::open(con, "rb")
     9. └─base::open.connection(con, "rb")
```

damianooldoni:
Yes, I read a lot months ago about mocks and cassettes and vcr package. I think I applied to some of our packages (I have to remember where I applied it).

However, I think you could also safely apply the `skip_on_cran()` function in your tests. 
See an example from rgbif package: https://github.com/ropensci/rgbif/blob/master/tests/testthat/test-downloads.R#L3

damianooldoni:
Now I remember where I applied mocks: it is for testing the `inborutils::setup_coding_club_session()`. See https://github.com/inbo/inborutils/blob/14729155612aff576e554d5091d8660547e2b2d0/tests/testthat/test-setup_codingclub_session.R#L23-L28 where I use `stub()` from the `mockery` package to "stub" the function and simulate some R basic functions.

peterdesmet:
Ok, need to fix this asap, package was removed: https://cran.r-project.org/web/packages/frictionless/index.html

peterdesmet:
Resolved in #116.

PR #116, opened 2022-11-09
Fix CRAN connection issues

damianooldoni:
This PR aims to solve #113, #114 and #115.

I used `testthat::skip_if_offline()` where needed in each of the test files. I preferred this skip function over `testthat::skip_on_cran()`. In this way a developer is free to test the package even if internet is (temporarily) unavailable.

About the vignette: I transformed some R chunks to markdown code where strictly needed. I know, this is not really elegant but I don't see really any other valid solution. Using package `mockery` and its function `stub()` will be way far less elegant than the solution proposed here.

### DIFF
--- a/tests/testthat/test-add_resource.R
+++ b/tests/testthat/test-add_resource.R
@@ -263,7 +263,6 @@ test_that("add_resource() can add resource from local, relative, absolute,
   expect_s3_class(read_resource(p, "compressed"), "tbl")
 
   # Remote
-  testthat::skip_if_offline()
   remote_path <- file.path(
     "https://github.com/frictionlessdata/frictionless-r",
     "raw/main/inst/extdata/deployments.csv"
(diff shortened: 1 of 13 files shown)


damianooldoni:
About bumping version, I didn't. I let you choose about it. Feel free to commit, of course.

tests/testthat/test-add_resource.R
```diff
@@ -85,6 +85,7 @@ test_that("add_resource() returns error on invalid or empty data frame", {
 })
 
 test_that("add_resource() returns error if CSV file cannot be found", {
+  testthat::skip_if_offline()
```
peterdesmet, line 88: @damianooldoni does this apply to the whole test or just the next assertion? Because currently (line 117) it is listed twice.

damianooldoni: You are right! it's a duplicate. If you remove your internet connection and run the test, you skip the very same number of tests (6). Fixed in https://github.com/frictionlessdata/frictionless-r/commit/74de9c90aff02a4be208d2ed696d76aa3df0bb06

peterdesmet: I also removed a number of other skip_if_online() tests that weren't needed.

damianooldoni:
@peterdesmet: I see you removed some "skips". Are you sure? You need internet connection for reading `example_package`.

PR #118, opened 2022-12-07
Use checklist package

peterdesmet:


damianooldoni (approved):
That was a massive amount of works, it seems. It seems that everything is ok. 👍

PR #122, opened 2023-01-26
 suppress messages from read_package

### LINKED ISSUE #121: suppress messages from read_package
## Description

`read_package()` is used in testthat tests within frictionless, where they are locally supressed [for example:](https://github.com/frictionlessdata/frictionless-r/blob/main/tests/testthat/test-read_package.R#L5)

```
p_local <- suppressMessages(read_package(p_path))
```

I also want to use `frictionless::read_package()` in some tests for a different package: etn. I'd like to propose adding an argument to the function to suppress messaging like:

> Please make sure you have the right to access data from this Data Package
    for your intended use.\nFollow applicable norms or requirements to credit
    the dataset and its authors.

I imagine this argument could be something like: 

```
silent = TRUE
silently = TRUE
supress_messages = TRUE
verbose = FALSE
show_messages = FALSE
```

I'll have a go at making this change myself, and adding an appropriate test + changing the tests to use the new argument instead of local versions of the `read_package()` function. 

## Proposed changes

- [ ] add argument to read_package()
- [ ] add coverage for new argument in test-read_package()
- [ ] change existing tests to make use of new argument

## Scope

I'm not proposing to suppress the assertions/checks made by `asserthat`, just the `message()` call ([link to line](https://github.com/frictionlessdata/frictionless-r/blob/421c22f8be948006c5fb89a822124fbb803dff12/R/read_package.R#L64))

PietrH:
## Motivation
Resolves #121

It would be convenient at times to suppress the rights/citation reminder in read_package(), including in the tests internal to frictionless-r. This pull request implements a new argument to read_package(), changes the internal testing to take advantage of this new argument, and adds a simple test to see if the new argument works. 

## Changes

- [x] decide on argument name
- [x] add argument to read_package()
- [x] add coverage for new argument in test-read_package()
- [x] change existing tests to make use of new argument
- [ ] Should there be a version dump? 
- [ ] News.md

### DIFF
--- a/R/read_package.R
+++ b/R/read_package.R
@@ -5,8 +5,8 @@
 #' that describes the Data Package metadata and its Data Resources.
 #'
 #' @param file Path or URL to a `datapackage.json` file.
-#' @param silently Do you want to suppress the rights/citations reminder?
-#'    This reminder can be disabled by setting `silently = TRUE`
+#' @param quietly Do you want to suppress the rights/citations reminder?
+#'    This reminder can be disabled by setting `quietly = TRUE`
 #' @return List describing a Data Package.
 #'   The function will add a custom property `directory` with the directory the
 #'   descriptor was read from.
@@ -25,7 +25,7 @@
 #'
 #' # List resources
 #' resources(package)
-read_package <- function(file = "datapackage.json", silently = FALSE) {
+read_package <- function(file = "datapackage.json", quietly = FALSE) {
   # Read file
   assertthat::assert_that(
     is.character(file),
@@ -65,7 +65,7 @@ read_package <- function(file = "datapackage.json", silently = FALSE) {
   }
 
   ## Optionally supress the message
-  if (!isTRUE(silently)) {
+  if (!isTRUE(quietly)) {
     message(msg)
   }
 
--- a/R/read_package.R
+++ b/R/read_package.R
@@ -48,13 +48,18 @@ read_package <- function(file = "datapackage.json", quiet = FALSE) {
   # Add directory
   descriptor$directory <- dirname(file) # Also works for URLs
 
-  # Inform user regarding rights/citations
+  # Inform user regarding rights/citations unless silenced
   msg <- glue::glue(
     "Please make sure you have the right to access data from this Data Package",
     "for your intended use.\nFollow applicable norms or requirements to credit",
     "the dataset and its authors.",
     .sep = " "
   )
+
+  if (!isTRUE(quiet)) {
+    message(msg)
+  }
+
   if (!is.null(descriptor$id)) {
     if (startsWith(descriptor$id, "http")) {
       msg <- glue::glue(
@@ -62,11 +67,6 @@ read_package <- function(file = "datapackage.json", quiet = FALSE) {
         .sep = "\n"
       )
     }
-  }
-
-  ## Optionally supress the message
-  if (!isTRUE(quiet)) {
-    message(msg)
   }
 
   descriptor
--- a/R/read_package.R
+++ b/R/read_package.R
@@ -5,7 +5,7 @@
 #' that describes the Data Package metadata and its Data Resources.
 #'
 #' @param file Path or URL to a `datapackage.json` file.
-#' @param quietly A logical. If `TRUE` no reminder about citations/rights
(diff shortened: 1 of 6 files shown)


R/read_package.R
```diff
@@ -5,6 +5,8 @@
 #' that describes the Data Package metadata and its Data Resources.
 #'
 #' @param file Path or URL to a `datapackage.json` file.
+#' @param quiet A logical. If `TRUE` no reminder about citations/rights
```
peterdesmet, line 8: Would simplify to:

> If `TRUE`, do not show message.

R/read_package.R
```diff
@@ -61,7 +63,11 @@ read_package <- function(file = "datapackage.json") {
       )
     }
   }
-  message(msg)
+
+  ## Optionally supress the message
+  if (!isTRUE(quiet)) {
```
peterdesmet, line 68: Why not:

```
if (!quiet) {
``` 

?

R/read_package.R
```diff
@@ -61,7 +63,11 @@ read_package <- function(file = "datapackage.json") {
       )
     }
   }
-  message(msg)
+
+  ## Optionally supress the message
```
peterdesmet, line 67: I would move the if statement up to line 49 and change comment to:

```
# Inform user regarding rights/citations unless silenced
```

tests/testthat/test-read_package.R
```diff
@@ -131,18 +129,27 @@ test_that("read_package() allows descriptor at absolute or relative parent
...
 test_that("read_package() allows YAML descriptor", {
   expect_true(
     check_package(
-      suppressMessages(read_package(test_path("data/valid_minimal.yml")))
+      read_package(test_path("data/valid_minimal.yml"), quiet = TRUE)
     )
   )
 })
+
+test_that(paste(
```
peterdesmet, line 148: I would not use paste for test descriptions, but simplify to:

> read_package() can silence message with quiet=TRUE

tests/testthat/test-read_package.R
```diff
@@ -131,18 +129,27 @@ test_that("read_package() allows descriptor at absolute or relative parent
...
+  "read_package() quiet argument allows to toggle",
+  "the rights/citation reminder"
+), {
+  p_path <- system.file("extdata", "datapackage.json", package = "frictionless")
+  expect_message(read_package(p_path, quiet = FALSE))
```
peterdesmet, line 153: Also include a test with the default:

```
expect_message(read_package(p_path))
```

peterdesmet (changes requested):

R/read_package.R
PietrH: from `help(isTRUE)`: 

> isTRUE(x) is the same as { is.logical(x) && length(x) == 1 && !is.na(x) && x }; isFALSE() is defined analogously. Consequently, if(isTRUE(cond)) may be preferable to if(cond) because of NAs.

This ensures that the message will be supressed only if the argument is exactly `TRUE` (or converted), not `1`, not `3`,  etc.. `rlang::is_true()` is even more strict I believe.

I wrote it more out of habit than anything else, I don't mind removing `isTRUE()` if you find it more readable that way. If you pass a string to quiet, `isTRUE()` will evaluate to `FALSE`, while `if("any string")` will result in an error: 

```r
Error in if ("any string") { : argument is not interpretable as logical
```

Any preference?

R/read_package.R
PietrH: It only silences the specific citation/rights message though, not any of the error messages from assertthat, for example:

```r
  assertthat::assert_that(
    is.character(file),
    msg = "`file` must be a path or URL to a `datapackage.json` file."
  )
```

I want to be clear because the effect is so specific. I'm ok with dropping "a logical"

tests/testthat/test-read_package.R
PietrH: done

tests/testthat/test-read_package.R
PietrH: done

R/read_package.R
PietrH: What about: 

> If `TRUE`, do not show message about access rights

PR #131, opened 2023-03-22
add readr::problems() to namespace

### LINKED ISSUE #129: Include `problems()` as part of namespace
Readr might return parsing issues:

```R
Warning message:                                                                                                                    
One or more parsing issues, call `problems()` on your data frame for details, e.g.:
  dat <- vroom(...)
  problems(dat) 
```

When the user does not have `readr` loaded, then trying `problems()` will return an error. This could be resolved by including it in the `frictionless` namespace, but I'm not sure how to do that exactly. A similar approach could be used for camtraptor.

PietrH:
Resolves #129 

For future reference:

```r 
usethis::use_package_doc(open = rlang::is_interactive())
usethis::use_import_from("readr","problems")
devtools::document()
```


PietrH:
ready for review

peterdesmet (changes requested):
When I install this package, I try ` problems(df)` I get:

> Error in problems(df) : could not find function "problems"

PietrH:
It helps to actually export functions instead of just adding them to the namespace, should be fixed now.

peterdesmet:
What command did you use to export it?

Also, we now get:

```
❯ checking for missing documentation entries ... WARNING
  Undocumented code objects:
    'problems'
  All user-level objects in a package should have documentation entries.
  See chapter 'Writing R documentation files' in the 'Writing R
  Extensions' manual.
```

PietrH:
`usethis::use_import_from()` followed by adding the export manually, I didn't run R CMD CHECK but clearly should have! 

I'm getting the idea this is not the best way to do this. Let me reevaluate and I'll let you know when I'm ready to revisit this review.

PietrH:
I've changed the implementation to use a `utils-*.R` script that only imports the namespace, and exports the function from readr. I've copied this mostly from how `usethis::use_pipe() `does it for `magrittr:::\\>%\\%`

The documentation just refers to readr, is this sufficient? It does pass `devtools::check()`

peterdesmet (approved):

R/utils-problems.R
```diff
@@ -3,10 +3,9 @@
 #' See [readr::problems()] for details.
 #'
 #' @name problems
-#' @rdname problems
-#' @keywords internal
-#' @export
 #' @importFrom readr problems
+#' @export
+#' @noRd
```
peterdesmet: @PietrH I was wrong. Not creating an `rd` will result in a CHECK warning. Fixed in 861d808. The "Help on topic 'problems' was found in the following packages:" is thus unavoidable.

PR #132, opened 2023-03-22
126 replace variable collapse with gluecollapse

### LINKED ISSUE #126: Replace variable_collapse with `glue::collapse()`
I found two cases:

https://github.com/frictionlessdata/frictionless-r/blob/421c22f8be948006c5fb89a822124fbb803dff12/R/check_schema.R#L60

https://github.com/frictionlessdata/frictionless-r/blob/421c22f8be948006c5fb89a822124fbb803dff12/R/get_resource.R#L25

Here's how to implement this better with glue

https://github.com/inbo/etn/blob/35e70f3d284cb64494313e667f238378564d494b/R/utils.R#L40-L43

PietrH:
Resolves #126 

Moves some objects that were only used for messaging inside glue syntax.


PietrH:
Ready for review

peterdesmet (changes requested):
This PR moves the variables to the `glue()` function. It does not use `glue_collapse()` though. Do you see a benefit in using this over ` paste()`?

PietrH:
In this case I don't see a benefit, but not a real cost either. So we can certainly use `glue_collapse()` if you prefer.

PR #136, opened 2023-04-05
134 Expand error message for col_select

PietrH:
Expanding the error message for `col_select` in `read_resource()`, as per example: 

> Can't find column(s) `animal-id`, `vis` in field names.
> ℹ Field names: `event-id`, `visible`, `timestamp`, `location-long`, `location-lat`, `bar:barometric-pressure`, `external-temperature`, `gps:dop`, `gps:satellite-count`, `gps-time-to-fix`, `ground-speed`, `heading`, `height-above-msl`, `import-marked-outlier`, `location-error-numerical`, `manually-marked-outlier`, `vertical-error-numerical`, `sensor-type`, `individual-taxon-canonical-name`, `tag-local-identifier`, `individual-local-identifier`, `study-name`

Also add tests

### renamed variable

the variable `field_names_collapse` does not actually refer to collapsed field names, but rather the elements of `col_select` that are missing from the schema, so I've renamed it to `missing_col_select_collapse`. I'm open to a better name. 

## Tests

We've had some trouble with testing error messages in the past, it seems changes in the error messaging aren't captured by `testthat::expect_error()`, for example adding a period at the end of the error, even when `fixed = TRUE`.

I also noticed the test did not fail when I added to the error message leaving the first part of the error intact: 84d0a45f337e19b17384825eb3488149555f01c9

So I've decided to switch to using snapshots wrapped in `cat()` to capture output, allowing errors (`testthat::expect_snapshot(error = TRUE)`). [testthat reference](https://testthat.r-lib.org/articles/snapshotting.html), [general reference](https://ro-che.info/articles/2017-12-04-golden-tests)

### Please consider

We could switch over all our `testthat::expect_error()` tests to using snapshots to increase defensiveness, at the possible cost of readability. So far these tests haven't caused any issues, but we should be aware that they might not capture all future changes in error messages.


PietrH:
It's unclear to me why my the action test-coverage fails when I [don't skip the snapshot tests for codecov.io](https://github.com/frictionlessdata/frictionless-r/actions/runs/4616332977), could you have a look at this @peterdesmet ?

```
Run covr::codecov()
  covr::codecov()
  shell: /usr/local/bin/Rscript {0}
  env:
    GITHUB_PAT: ***
    R_LIBS_USER: /Users/runner/work/_temp/Library
    TZ: UTC
    _R_CHECK_SYSTEM_CLOCK_: FALSE
    NOT_CRAN: true
Error: Error in file(con, "r") : cannot open the connection
Calls: <Anonymous> -> package_coverage -> add_hooks -> readLines -> file
In addition: Warning messages:
1: In utils::install.packages(repos = NULL, lib = install_path, pkg$path,  :
  installation of package ‘/Users/runner/work/frictionless-r/frictionless-r’ had non-zero exit status
2: In file(con, "r") :
  cannot open file '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/Rtmpc7uIY2/R_LIBSe61d4dce65/frictionless/R/frictionless': No such file or directory
Execution halted
Error: Process completed with exit code 1.
```

peterdesmet:
See my last 3 commits:

- The entire error message _is_ checked with `fixed=TRUE`. I use `paste(sep = "\n")` to break the multiple lines. Some of the lines extend 80 characters, but left it like that for readability. I removed the snapshot approach.
- I remove the `expect_no_error()` test, as that is covered in tests above.
- I renamed `schema_names_collapse` back to `field_names_collapse`. I'm fine with `missing_col_select_collapse`

PietrH:
Thanks for the input, looks good to me.

PR #137, opened 2023-06-14
add explicit default simplifyVector=T on fromJSON call

mpadge:
Hi @peterdesmet, this PR is cosmetic only, and just makes a default parameter value explicit. I actually needed this to be `FALSE` for some of my deposits work, so followed through why you use the default `TRUE` here. Your reason for that now makes sense to me, but having it explicitly here would have helped my initial reading of the code.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -132,7 +132,7 @@ read_descriptor <- function(x, directory = NULL, safe = FALSE) {
       x <- yaml::yaml.load_file(x)
     } else {
       # Default to jsonlite: better error messages for non .json files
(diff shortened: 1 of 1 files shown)


peterdesmet:
Thanks for diving so deep into the code. Will you be able to work around it for your use case where you want this `FALSE`?

mpadge:
Thanks for asking - yeah, it wasn't too hard to write work-arounds in most cases. Only once did i have to double-read, first via frictionless, then again with `read_json(simplifyVector = FALSE)`; all other cases were either/or. All good!

PR #138, opened 2023-06-14
use scalar logical operators

mpadge:
I noticed some of these while examining #137. A minor improvement with  no effect on pkg behaviour.


peterdesmet:
Thanks for the contribution!

Issue #140, opened 2023-06-16
Support additional resource properties with `add_resource(...)`


peterdesmet:
`add_resource()` will [add](https://github.com/frictionlessdata/frictionless-r/blob/b0a9fe34b26fe120309899503bba67cffcb710ff/R/add_resource.R#L116) the following properties to resources:

- name
- data or path
- profile
- schema
- format (when path is used)
- mediatype (when path is used)
- encoding (when path is used)
- dialect (when path is used + dialect is not default)

But resources can have [additional properties](https://specs.frictionlessdata.io/data-resource/#optional-properties):

- `title`
- `description`
- `bytes`
- `hash`
- `licenses`

Rather than adding all those as extra parameters in `add_resource()`, they could be supported with ellipsis `...`. See the suggestion by @hansvancalster at https://github.com/frictionlessdata/frictionless-r/issues/48#issuecomment-1594399115

peterdesmet:
Note that there are only 5 unsupported additional properties, so we could be defensive and hardcode those. On the other hand, the Frictionless specs do allow to expand with your own properties, so I still think that `...` is a good approach. But it will be up to the user then, that e.g. adding a property `hash` or `bytes` meets the requirements of the definition.

I would return a message if custom properties are defined:

> The following custom properties were provided:
> ✖ `path`: ignored, is set by the function.
> ℹ `sources`: added, should meet requirements defined at https://specs.frictionlessdata.io/data-resource/#optional-properties.
> ✅ `blabla`: added.

That message could be silenced by `suppressMessages()`

peterdesmet:
Note to self: implementation in #141 returns:


``` r
library(frictionless)
p <- create_package()
p1 <- add_resource(p, "mtcars", mtcars)
p2 <- add_resource(p, "mtcars", mtcars, foo = "bar")
#> The following custom properties were provided:
#>   foo: added.
p3 <- add_resource(p, "mtcars", mtcars, source = "test")
#> The following custom properties were provided:
#>   source: added.
p4 <- add_resource(p, "mtcars", mtcars, path = "test")
#> Error: path are passed to ... but are set internally. Please remove them.
```

<sup>Created on 2023-06-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

peterdesmet:
On second inspection, I think it is better:

- Not to return a message. This is unnecessarily verbose for a functionality that will mostly be used by other functions.
[...]

PR #141, opened 2023-06-16
update for add_resource() to allow optional metadata

hansvancalster:
This PR implements #140


peterdesmet (approved):
Thanks for the initial implementation @hansvancalster. I've made substantial changes in line with https://github.com/frictionlessdata/frictionless-r/issues/140#issuecomment-1597256761. The function no longer returns a `message()`, simplifying implementation.

PR #146, opened 2023-08-03
127 compare header and schema

PietrH:


PietrH:
`check_schema()` fails on case mismatch between schema and data. To replicate: 

```R
# create package with the wrong case in the schema of observations
  wrong_case_in_schema_pkg <- example_package
  ## change case
  purrr::pluck(wrong_case_in_schema_pkg, "resources", 2, "schema", "fields") <-
    purrr::chuck(wrong_case_in_schema_pkg, "resources", 2, "schema", "fields") %>%
    purrr::map2_chr(
      c(TRUE, TRUE, FALSE, FALSE, FALSE, TRUE, TRUE), #set some to upper, some to lower case
      ~ ifelse(.y,
        toupper(purrr::chuck(.x, "name")),
        tolower(purrr::chuck(.x, "name"))
      )
    )
read_resource(wrong_case_in_schema_pkg, "observations")
```

PR #152, opened 2023-09-18
🔗 Fix link to `datapackages-r`

nepito:
The link to `datapackages.r` is broken. 

This new link refers to a repository `datapackages-r` from the **frictionlessdata** organization.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -93,7 +93,7 @@ For more functionality, see [get started](https://docs.ropensci.org/frictionless
 
 ## frictionless vs datapackage.r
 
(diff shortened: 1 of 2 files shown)


peterdesmet:
Thanks! While `https://cran.r-project.org/package=datapackage.r` is the canonical way to refer to the package, the GitHub link is more useful to users as it provides documentation and instructions to install from GitHub. The name should remain `datapackage.r` (that is the official name as defined in https://github.com/frictionlessdata/datapackage-r/blob/master/DESCRIPTION#L1).

Will update to use the new link (and same name) in `README.Rmd` (the source of `README.md`)

Issue #154, opened 2023-10-09
Pretty print package objects


khusmann:
Thanks for all your work on this package! I really appreciate the simplicity & tidyverse principles. I hope you don't mind if I jump in!

Presently, printing loaded packages in the console simply dumps their list representation (which can span many pages in larger packages). I think it would be really nice for packages to have a pretty-printed summary by default instead. This is handy for interactive sessions, gives the package a really polished look, and I think it would help with adoption efforts by making things a little less intimidating for new users. 

I'll submit a pull request in a sec with a candidate implementation...

peterdesmet:
I agree! Makes me also wonder if we should move the console message one currently gets when using `read_package()` to the print function. The message is a bit annoying for packages relying on this package. Ping @damianooldoni.

khusmann:
> Makes me also wonder if we should move the console message one currently gets when using read_package() to the print function.

Oh, I like this idea! I definitely found the default loading message a little annoying for exactly that reason, and was planning to request an option to disable it. It feels a lot more natural in a `print` function.

peterdesmet:
Question: how would a user get the default (non-pretty) print? This might especially be useful for developers building on top of frictionless. And what would the result be of:

```R
str(package, list.len = 3)
```

khusmann:
> how would a user get the default (non-pretty) print?

Ah, good thought. As a user I would expect to do something like `as.list(example_package)` to remove the datapackage class. If you like that approach lmk and I can add.

> `> str(pkg, list.len=3)`

```
List of 8
 $ name     : chr "example_package"
 $ id       : chr "115f49c1-8603-463e-a908-68de98327266"
 $ created  : chr "2021-03-02T17:22:33Z"
  [list output truncated]
 - attr(*, "class")= chr "datapackage"
```

peterdesmet:
`as.list()` sounds like a good approach. @damianooldoni would you agree?

peterdesmet:
This is now implemented in #184. The `message()` is removed from `read_package()`. To get the default print, `unclass()` is suggested, not `as.list()`.

PR #155, opened 2023-10-09
add a print function for data packages

khusmann:
Here's a candidate implementation for #154 

Might be worth experimenting with other designs / info views as well, but this is a start!

It's been a while since I've hacked on an R package, please bear with me as I get back up to speed... (you can let me know as we go if you'd like me to adjust my style)


khusmann:
For reference, here's an example output:

```
read_package(system.file("extdata", "datapackage.json", package="frictionless"))
```

Output:

```
# A frictionless data package: 'aces2-main'
# 4 available tabular data resource(s): 

                                name fields description
1                    tables-teachers     83           ~
2                    tables-students    147           ~
3          tables-students_occasions      2           ~
4 tables-students_teachers_occasions    422           ~

# i Use `read_resource()` to load tabular data resource
```

peterdesmet:
Nice! I’m currently out of office, so it might take me a while to review. One thing I noticed is that you assign a custom class `fls_pkg`. Can you rename that class (and associated functions) to `datapackage`? I think that is more intuitive and it aligns with the `datapackage.json` convention.

khusmann:
> Can you rename that class (and associated functions) to datapackage

Sure thing! That was my first instinct as well, but I was a little worried there might be name collisions with other packages. I just checked datapackage.r and that won't be a problem, so that's good. Can I search CRAN for class names? Or should we just go for it?

peterdesmet:
I would just go for it!

R/create_package.R
```diff
@@ -16,9 +16,12 @@
 #' package <- create_package()
 #' str(package)
 create_package <- function() {
-  list(
-    profile = "tabular-data-package",
-    resources = list(),
-    directory = "." # Current directory
+  structure(
```
peterdesmet, line 19: There is old code that used to add a `datapackage` class:

https://github.com/frictionlessdata/frictionless-r/blob/01e383f81272ba32fe33ca933896e1e5a7a5f77f/R/create_package.R#L26-L28

It first creates the list, and only then assigns the class (no `structure()`). It also stores any already existing classes, rather than hardcoding `list`. Can you use that approach instead?

khusmann: Yup! Also reworked create_package a bit so that it creates from a descriptor, so we can then use it as a factory in read_package

peterdesmet:
@khusmann see https://github.com/khusmann/frictionless-r/pull/1, which contains changes I made to this PR. Can you review and accept? That will allow us to pull those changes into this PR and then `main`.

Issue #157, opened 2023-10-10
Automatically load "datapackage.json" when supplied a path to a folder


khusmann:
A lot of frictionless implementations automatically load datapackage.json when reading a directory. It'd be nice to go

```
read_package("/path/to/packagedir")
```

instead of always

```
read_package("/path/to/packagedir/datapackage.json")
```

peterdesmet:
Would be useful, but it is a bit more complex than:

- Check if the provided path (or URL) ends with `datapackage.json`
- If not, append `datapackage.json` with `file.path()`
- Return error if that file could not be found

`datapackage.yml` and `datapackage.yaml` files are also valid, so we need to check if the provided file path has this. If not, we'll have to assume a `datapackage.json` file, potentially missing a `datapackage.yml` file. Maybe we could check for those files as well before reporting an error.

I think that is why I initially chose the verbose approach, especially since tab completion when writing the path immediately provides feedback on whether a file is present.

peterdesmet:
I'm curious to see how other frictionless software tackles this.

khusmann:
Ah, makes sense! I was wondering about that yaml stuff -- I saw yaml export functions in frictionless-py but so far have not seen a datapackage.yaml in the wild, so I was running with the assumption that the datapackage.json was the defacto standard.

The collections in the datahub were initially confusing to me to get working with `frictionless-r` because there was no direct link to the datapackage.json in their file listing ([here, for example](https://datahub.io/core/exchange-rates)). The default behavior of their [data-cli tool](https://datahub.io/tools) points to the root URL of the package though, and I found adding /datapackage.json did the trick.

I think it's nice (for new users especially) to be able to treat the datapackage as a sort of opaque blob they can load resources from (like tabs in an excel file), without needing to think about the internal structure -- it also facilitates distributing packages as self-contained zip files.

peterdesmet:
@khusmann Thanks for investigating. `datapackage.json` is the only valid format according to the specs, but `yml/yaml` is supported by frictionless-py and it was requested and implemented as a feature for frictionless-r. I think for guessing a file, it's fine to follow frictionless-py (and the specs) and only look for a `datapackage.json`. I'll try to get your PR included in the next version.

peterdesmet:
Update:

[...]

PR #158, opened 2023-10-12
When read_package is supplied a path to a directory, try datapackage.{json|yaml|yml}

khusmann:
A possible way to implement #157 . It sounds like you're plan is to let the idea simmer to see how other implementations approach it, but figured I'd submit what I have so you'd have it in your back pocket. Notes:

When the data package path, is an URL, it does not attempt to try the candidate datapackage resource filenames. I figured that with url redirects, etc, a URL without a ".{json|yaml|yml}$" could still exist (no 404) and be pointing to a valid resource file. And we can't really rely on content-type... so maybe better to not mess with. Another reason to not mess with urls is that this matches the behavior of `frictionless-py`. (Although it appears that [data-cli from datahub ](https://datahub.io/tools)  will happily add /package.json to urls, as I mentioned earlier...)

In my (brief) testing, the `frictionless-py` cli will only try "datapackage.json" as a candidate file to load when pointed to a directory... it does not discover datapackage.yaml or datapackage.yml. If we wanted to match this behavior, it'd only require editing the autoload candidate list [here](https://github.com/khusmann/frictionless-r/blob/d757750fc4c411b1af94505cb90b28f9fbcd8e18/R/read_package.R#L35).


peterdesmet:
I have simplified this PR and added tests.

1. When a directory rather than a file is provided, it will only look for `datapackage.json` (not yaml or yml)
2. I actually want this functionality for URLs, not only for local directories, so that the example posted in #157 (https://datahub.io/core/exchange-rates) works.
3. This functionality at the level of `read_package()`, keeping `read_descriptor()` and `check_path()` simple. The former is also used for getting Table Schema.
4. The implementation is now very simple: if `file` ends with `/` assume a directory and add `datapackage.json`. Annoyingly, https://datahub.io/core/exchange-rates won't work, but https://datahub.io/core/exchange-rates/ will. I looked for some alternative implementations to determine if `file` is a directory or url (without`datapackage.json`) but:
  - `dir.name()` doesn't work well for this (e.g. `dirname("my_directory")` returns `.`)
  - `file.info()$isdir` only works for local directories that exist.
  - I didn't want to look for a range of file names or extensions at the end, otherwise `file = "datapackage.csv"` would return "Can't find file at 'datapackage.csv/datapackage.json'"
 
Alternatives to support `mydir/` as well as `mydir` are appreciated.

peterdesmet:
Closing, see https://github.com/frictionlessdata/frictionless-r/issues/157#issuecomment-2309803768

PR #182, opened 2024-03-15
Don't hardwrap text in news

peterdesmet:
I notice in other packages (and style guides) that NEWS.md is hard-wrapped at 80 chars and has lots of spacing between bullet items. I find this annoying:

- It renders ok (wrapping and spacing is ignored) on GitHub and pkgdown website.
- But the text is also often copy/pasted for GitHub and Zenodo releases, where the hard-wrapping is not ignored.

In addition, this is a pure Markdown file, I rather use the Markdown conventions over the R conventions.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,89 +1,57 @@
 # frictionless (development version)
 
-- `add_resource()` now supports adding additional resource properties via the 
-  `...` argument.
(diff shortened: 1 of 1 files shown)


damianooldoni (approved):
I totally agree with your choice and the motivation of it.

PR #183, opened 2024-03-15
Avoid #156

### LINKED ISSUE #156: Documentation build error
FYI, when running `devtools::document()`, the build fails with the following warning/error when `rlist` is not installed:

```
ℹ Updating frictionless documentation
ℹ Loading frictionless
Warning message:
[utils.R:36] @description refers to unavailable topic rlist::list.clean
Caused by error in `find.package()`:
! there is no package called ‘rlist’ 
```

Not sure if there's any way around this... is it possible to make this reference valid without installing rlist?

Alternatively, you might consider just using `purrr::keep`, `purrr::discard`, and `purr:compact`, I think they do basically the same thing and you're already requiring purrr.

In any case, it's very low priority because it's an internal function & only affect developers -- just wanted to give you the heads up.

peterdesmet:
By using backticks, Roxygen will just create:

```
\code{}
```

Not:

```
\code{\link[]{}}
```

And not generate an error if rlist is not installed by the contributing developer. Fix #156

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -36,7 +36,7 @@ unique_sorted <- function(x) {
 #' Removes all elements from a list that meet a criterion function, e.g.
 #' `is.null(x)` for empty elements.
 #' Removal can be recursive to guarantee elements are removed at any level.
-#' Function is copied and adapted from [rlist::list.clean()] (MIT licensed), to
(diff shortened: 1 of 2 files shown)


PietrH:
Rerunning failed jobs

PietrH (approved):

Issue #185, opened 2024-03-19
Public use of `check_package()`


sangovae:
It would be usefull if the `check_package()` function is public so we can use it in the camtrapdp package.

PR #187, opened 2024-03-19
Add `datapackage` class, update `create_package()` and `check_package()`

### LINKED ISSUE #184: Add class to `package` object
To better recognize a `package` object, e.g. for pretty print (#154), it would be good to extend the list class with a custom class `datapackage`.

- [x] Add class `datapackage`
- [x] Add the class with `create_package()`
- [x] Allow `create_package()` to start from an existing list and use it in `read_package()`
- [x] Add tests for `create_package()`
  - [x] Including for existing lists
- [x] Test for class in `check_package()`, **warn** if not present
  - [x] Maybe provide more specific error message on what is invalid
- [x] Add test for `check_package()`
- [x] Make `check_package()` public, so other packages can use it

peterdesmet:
Fix #184, #185 

- Add `datapackage` class to `package` and describe it throughout as a "Data Package object"
- Allow to provide an existing descriptor to `create_package()` to make it a Data Package object. Use it in `read_package()`
- Always run a `check_package()` at the end of `create_package()`
- Make `check_package()` public and provide more meaningful error messages. Warn if `datapackage` class is missing, so that packages using frictionless as a dependency don't break.
- #188


R/read_package.R
```diff
@@ -64,5 +61,5 @@ read_package <- function(file = "datapackage.json") {
   }
   cli::cli_inform(message, class = "frictionless_message_usage_rights")
 
-  descriptor
+  create_package(descriptor)
```
PietrH, line 64: Why not:
```suggestion
return(create_package(descriptor))
```

Like for `create_package()`?

R/read_resource.R
```diff
@@ -10,13 +10,13 @@
...
-#' @param package List describing a Data Package, created with [read_package()]
-#'   or [create_package()].
+#' @param package Data Package object, created with [read_package()] or
+#'   [create_package()].
 #' @param resource_name Name of the Data Resource.
 #' @param col_select Character vector of the columns to include in the result,
 #'   in the order provided.
 #'   Selecting columns can improve read speed.
-#' @return A [tibble()] data frame with the Data Resource's tabular data.
+#' @return [tibble()] data frame with the Data Resource's tabular data.
```
PietrH, line 19:
```suggestion
#' @return [dplyr::tibble()] data frame with the Data Resource's tabular data.
```

R/create_package.R
```diff
@@ -1,24 +1,46 @@
...
+  descriptor$directory <- replace_null(descriptor$directory, ".") # Current dir
+
+  # Add datapackage class
+  if (!"datapackage" %in% class(descriptor)) {
+    class(descriptor) <- c("datapackage", class(descriptor))
+  }
+
+  # Check that created package is valid
+  check_package(descriptor)
+  return(descriptor)
```
PietrH, line 45:
```suggestion
  check_package(descriptor)
  
  return(descriptor)
```

I'd vertically split `check_package()` from `return()` because the comment above is not about `return()`

PietrH (changes requested):
Tiny suggestions, nothing that needs to be changed per se.

Very nicely done, very clear and easy to review. 

- I've found another #156 and fixed it. 4f228aced31da714d75d2c025b262691fdc2a501
- Currently in the differences with datapackage.r section of the [readme](https://github.com/frictionlessdata/frictionless-r#frictionless-vs-datapackager) we mention the fact that datapackage.r outputs a `Package` object as a major difference. This is no longer true.

R/read_package.R
peterdesmet: I mainly want to avoid "naked" variables. So I don't see a need to double wrap here.

peterdesmet (commented):
@PietrH implemented most of the suggested changes. Can you review again?

PR #189, opened 2024-03-23
Have `check_` functions return invisibly rather than `TRUE`

peterdesmet:
@damianooldoni I agree with what you said, there isn't a good argument to have these return TRUE. Should be updated in camtrapdp as well.


damianooldoni (approved):
Thanks for thinking about this! 
I agree with all changes. I also checked and tested the package locally. Documentation is also compiled correctly. Top work 👍

damianooldoni:
I approved the PR, but notice there are conflicts to solve before merging.

PR #192, opened 2024-03-25
Remove `replace_null()` in favour of rlang's `%||%`

### LINKED ISSUE #175: Consider using `rlang::%||%` over `replace_null()`
See https://masalmon.eu/2023/06/06/basic-patterns/ There is an rlang function `%||%` that has same functionality as `replace_null()`.

Advantages:

- Existing functionality

Disadvantages:

- New dependency... although, seems it is already used in

https://github.com/frictionlessdata/frictionless-r/blob/f7f3fe3b2077843113538428c51a1f2d9a736b9a/R/read_resource.R#L379-L381

- More clunky to call with namespace `rlang::%||%`, would have to revert to `%||%`

peterdesmet:
Fix #175. Requires `rlang` as dependency, but users will already have it installed via the other dependencies of frictionless.

Could be argued if it is more readable, but it's a convention at least and it makes for more concise code.


R/utils.R
```diff
@@ -28,7 +13,7 @@ unique_sorted <- function(x) {
   # 1 2 3
   values <- names(sort(table(x), decreasing = TRUE))
   # Return empty char vector if all values in x where NA, resulting in NULL
```
damianooldoni, line 15: "are" instead of "where"?

damianooldoni (approved):
I agree with the use of `%||%` and importing rlang. I do not really agree that it's more readable than `replace_null()`, but the pros are definitely more than the cons.

Package also checked and tested locally. Everything works perfectly. Documentation is compiled correctly.

R/utils.R
peterdesmet: Correct, fixed in 63eed65

PR #199, opened 2024-03-27
Precompile vignette

### LINKED ISSUE #133: Mention `deposits` package in Vignette
A very promising [`deposits` package](https://github.com/ropenscilabs/deposits) is currently being developed. It aids users with submitting (and retrieving) research products to (and from) Zenodo and Figshare. It relies on the `frictionless` R package to document items with metadata.

We currently conclude our vignette with:

https://github.com/frictionlessdata/frictionless-r/blob/888bc97e62269704bfeaf52f3ebc3a084d3009d3/vignettes/frictionless.Rmd#L258-L262

That seems like a good place to mention the `deposits` package. My main question is timing, as `deposits` is currently indicated as Work In Progress. @mpadge any advice?

### LINKED ISSUE #194: Warning in vignette example
See https://docs.ropensci.org/frictionless/articles/frictionless.html#write-a-data-package, which currently has:

```
#> Warning: `package` must be a Data Package object.
#> ✖ `package` is missing a "datapackage" class.
#> ℹ Create a valid Data Package object with `read_package()` or
#>   `create_package()`.
```

peterdesmet:
The vignette used to [fail on CRAN sometimes](https://github.com/frictionlessdata/frictionless-r/issues/113), because it depends on zenodo.org, which can have service failures. This was solved by replacing code blocks with rendered code blocks in https://github.com/frictionlessdata/frictionless-r/commit/b9f83edd84276a43e2a2db462fe88c805cc5b780. The downside is that those code blocks (and their output) are never re-rendered (or had to be updated manually).

This PR:

- Implements an approach suggested in https://doi.org/10.59350/x0a46-gk734:
  - Adds a `frictionless.Rmd.orig` file. It is where the actual vignette lives, but is ignored by CRAN or R CMD Check.
  - The `frictionless.Rmd` is the knitted version of `frictionless.Rmd.orig`. It can be created by running `precompile.R`.
  - The advantage is that the vignette is not run on CRAN or R CMD Check, saving resources, time and potential errors. The 7K downloads of https://zenodo.org/records/10053702 were likely caused by our GitHub actions, so it's good to reduce that. I don't mind that the vignette isn't "tested", we have tests for that. See this [quote](https://blog.r-hub.io/2020/06/03/vignettes/#hey-what-about-testing-and-reproducibility) in agreement.
  - The downside is that one needs to manually run `precompile.R`. This should then be done before a release.
- Mentions deposits. Fix #133.
- Provides a temp solution for the write_package() warnings. Fix #194.
- Uses the latest version of the O_WESTERSCHELDE dataset
- Removes the vignette date


peterdesmet:
An alternative to all this would be to create an [article](https://blog.r-hub.io/2020/06/03/vignettes/#how-to-include-a-compute-intensive--authentication-dependent-vignette) (that only shows on the pkgdown website) instead of a vignette. Downsides:

- Wouldn't save resources
- An introduction vignette is required by rOpenSci

damianooldoni:
I agree with choosing the approach suggested in https://doi.org/10.59350/x0a46-gk734 instead of creating articles. Even if we would have more than one vignette, I would still not create articles for frictionless. A review will follow.

vignettes/frictionless.Rmd
```diff
@@ -72,48 +70,51 @@ str(package, list.len = 3)
...
-#>  2 14256075763 TRUE    2018-05-25 16:16:41            4.25           51.3
-#>  3 14256075764 TRUE    2018-05-25 16:21:29            4.25           51.3
-#>  4 14256075765 TRUE    2018-05-25 16:26:28            4.25           51.3
-#>  5 14256075766 TRUE    2018-05-25 16:31:21            4.25           51.3
-#>  6 14256075767 TRUE    2018-05-25 16:36:09            4.25           51.3
-#>  7 14256075768 TRUE    2018-05-25 16:40:57            4.25           51.3
-#>  8 14256075769 TRUE    2018-05-25 16:45:55            4.25           51.3
-#>  9 14256075770 TRUE    2018-05-25 16:50:49            4.25           51.3
-#> 10 14256075771 TRUE    2018-05-25 16:55:36            4.25           51.3
+#>     `event-id` visible timestamp           `location-long` `location-lat` bar:barometric-pressur…¹
```
damianooldoni, line 91: If I knit `vignettes/frictionless.Rmd.orig` via `compile.R` I get a different output. For example, `bar:barometric-pressure` instead of `bar:barometric-pressur…¹`  in this line.
I noticed that everytime I change the window size and/or the font size I get a new result. Is this a Windows specific behavior? Can you check on your MacOS, please? It's a quite annoying behavior.

vignettes/frictionless.Rmd
```diff
@@ -256,24 +353,28 @@ Note that in the above steps you started a Data Package from scratch with `creat
...
 ```
 
 The directory will contain four files: the descriptor `datapackage.json`, one CSV file containing the data for the resource `iris` and two CSV files containing the data for the resource `observations`.
 
+
 ```r
 list.files("my_directory")
-#> [1] "datapackage.json"   "iris.csv"           "observations_1.csv"
-#> [4] "observations_2.csv"
+#> [1] "datapackage.json"   "iris.csv"           "observations_1.csv" "observations_2.csv"
```
damianooldoni, line 369: If I run `compile.R` I get `character(0)`.

damianooldoni (approved):
Nice work and I learned a lot too. Approved changes. Tests, checks and documentation are ok.

Just a note: I found an annoying behavior while knitting `frictionless.Rmd.orig` vignette (via `compile.R`): the generated `.Rmd` output depends on local settings such as window size and font size. Even if the same person knit the vignette, let's say Peter, the "future-Peter" will have to be careful that the same local settings are used.
@peterdesmet: I did a fast search but didn't find a real solution for this.

vignettes/frictionless.Rmd
peterdesmet: Caused by incorrect function call, fixed in d781c83.

vignettes/frictionless.Rmd
peterdesmet: Note: this `option` will reset itself if you close the session, so it doesn't affect the developer that much.

PR #230, opened 2024-07-09
write_package() does not write to "." by default anymore

### LINKED ISSUE #205: `write_package()` should not write to `"."` by default
This is a violation of CRAN policies, see https://github.com/inbo/camtrapdp/pull/79

- [x] Provide no default for `directory`
- [x] Verify tests still run
- [x] Verify vignette needs update

sannegovaert:
fix #205


peterdesmet (approved):
Nice! I guess an extra test is not needed. Please update the NEWS.md, cf:

> * `write_package()` no longer writes to `"."` by default, since this is not allowed by CRAN policies. The user needs to explicitly define a directory (#205).

PR #240, opened 2024-07-09
Setting minimum R dep to R4.0

PietrH:


PietrH:
This will also bump the minimum requirements of camtrapdp and camtraptor

peterdesmet:
Closed in favour of #245.

PR #242, opened 2024-07-10
Replace magrittr pipes with base pipes

PietrH:
After #240, we can use base pipes in our vignettes and examples

What is `frictionless.Rmd.orig` used for?


peterdesmet:
Closing this without merging:

- Base pipes are only introduced since R4.1.0 (and we planned to require R4.0.0): https://www.tidyverse.org/blog/2023/04/base-vs-magrittr-pipe/#pipes
- We will support R3.6.0

Sorry @PietrH 😬

PR #229, opened 2024-07-08
Encode null as NULL and NA as string

peterdesmet:
Fix #203. This PR:

- [x] Needs to be extended with a new test
- [x] Verified if current tests don't fail

PR #243, opened 2024-07-10
add null value to datapackage.json

sannegovaert:
To be able to update example_dataset in https://github.com/frictionlessdata/frictionless-r/pull/229


peterdesmet (approved):

peterdesmet:
Tests are expected to fail for this PR (since the example package is no longer the same as the rda file). This is to be resolved in #229

PR #248, opened 2024-07-12
Example package function

### LINKED ISSUE #114: Do not require internet connection for `example_package`

peterdesmet:
Fix #114.

## What changed

> The included dataset `example_package` is removed in favour of the function `example_package()`. This function allows to reproducibly provide a _local Data Package_, while before it needed to be a remote package. The `observations` resource was also changed from a remote to a local resource, allowing the entire example Data Package to be read locally. Examples and tests were updated (#114).

Note: tests reassign a `path` to a URL to test remote reading. Some tests require one of two paths to be reassigned (mixed local/remote). I left that like it is and did not implement a `remote` argument in `example_package()`, as it wouldn't solve that last use case.

## Tests

Thanks to this change, 19 tests can now run without internet, and no longer require `skip_if_offline()`. Some tests currently failed without internet (unrelated to example package), which I fixed too.

test | fails if offline | skipped | skipped with example_package() | skipped with local observations
--- | --- | --- | --- | ---
add_resource | 0 | 2 | 2 | 2
check_data* | 0 | 0 | 0 | 0
check_package | 0 | 0 | 0 | 0
check_path* | 2 (unrelated to example_package) | 0 | 2 | 2
check_schema | 0 | 1 | 0 | 0
create_package* | 0 | 0 | 0 | 0
create_schema* | 0 | 0 | 0 | 0
print | 0 | 0 | 0 | 0
read_package* | 0 | 2 | 2 | 2
read_resource | 0 | 13 | **5** | 5
remove_resource | 0 | 3 | **0** | 0
resources | 0 | 1 | 1 | **0**
utils* | 0 | 0 | 0 | 0
write_package | 1 (unrelated to example_package) | 14 | 15 | **8**


sannegovaert (approved):
Very nice!

PR #253, opened 2024-08-19
Use a deviating dialect for one of the resources in `example_package`

### LINKED ISSUE #250: Use a deviating dialect for one of the resources in `example_package`
To indicate that a `datapackage.json` makes use of v2 Table Dialect features, one needs to define:

```json
"dialect": {
  "$schema": "https://datapackage.org/profiles/2.0/tabledialect.json"
}
```

Currently, none of the resources in `example_package` define a dialect (they all use defaults). As a result, it will also use the default for the not-defined `$schema`, which is v1, thus blocking any testing of v2 features.

Solution: define the dialect for one of the resources in the `example_package`. I suggest "observations", with e.g. tab-based delimiter.

- [x] Add a `"dialect": {"delimiter": "\t"}` to the observations resource in `datapackage.json`
- [x] Update the files to use tabs
- [x] Rename the files to `observations_1.tsv` and `observations_2.tsv`
- [x] Update reference to these files in tests
- [x] Run tests
- [x] Update reference to these files in examples and vignettes (especially comments)

sannegovaert:
fix #250 

- Some tests failed because they use a link at Github. This can only be tested after merging?
- There is an error at line 88 of `vignettes/frictionless.Rmd` which seems unrelated to #250 ?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
 * `read_resource()` now returns error if both `path` and `data` are provided (#143).
 * `write_package()` no longer writes to `"."` by default, since this is not allowed by CRAN policies. The user needs to explicitly define a directory (#205).
 * `write_package()` now writes incoming `null` values back to `NULL` in `datapackage.json`, rather than empty an empty lists. Properties that are assigned `NA` and `NULL` by the user, remain being written as `null` and removed respectively (#203).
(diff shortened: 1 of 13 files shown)


peterdesmet (approved):
Excellent!

PR #258, opened 2024-08-27
Create a v2 `example_package()`

### LINKED ISSUE #249: Create a v2 example_package
To support tests for Data Package v2, while still maintaining v1, we will need different versions of the example package. That should be pretty easy with the `example_package()` function

## Step 1

- [x] Fix #250 first
- [x] Move content of [inst/extdata](https://github.com/frictionlessdata/frictionless-r/tree/main/inst/extdata) to `inst/extdata/v1`
- [x] Update `example_package()` to read from this directory
- [x] Update any paths to these files in the tests
- [x] Run tests. These will likely fail for remote URLs until things are updated on `main`
- [ ] Push to `main`, run tests again
- [x] Make sure frictionless v1.2.0 still works (so we don't repeat #256). Normally it should, since we're reading locally from data shipped in inst with the package.

## Step 2

- [x] Create a `version` parameter in `example_package()`
- [x] Based on the version, read from `inst/extdata/<version>/`
- [x] Default to `v1`, so all tests remain to work

## Step 3

- [x] Create a v2 example package at `inst/extdata/v2` @peterdesmet
- [x] Create `test-example_package.R` to test this parameter, similar to:

https://github.com/frictionlessdata/frictionless-r/blob/2f80fed68fbfa1013a117082391d21a7cbdcac44/tests/testthat/test-read_resource.R#L287-L293

sannegovaert:
fix #249


peterdesmet:
> Are you making a v2 example_package?

Depends if you want to try yourself. :-) The guiding document is https://datapackage.org/overview/changelog/

R/example_package.R
```diff
@@ -9,12 +9,24 @@
...
+#' https://datapackage.org/blog/2024-06-26-v2-release/) standard that this
+#' example Data Package uses. Currently only version 1 is supported.
 #' @return A Data Package object, see [create_package()].
 #' @family sample data
 #' @export
 #' @examples
 #' example_package()
-example_package <- function() {
-  path <- system.file("extdata", "datapackage.json", package = "frictionless")
+example_package <- function(version = 1) {
```
peterdesmet, line 20: Version should be a string value `"1.0"` Use camtrapdp logic to check for version compatibility:

https://github.com/inbo/camtrapdp/blob/573e4b822718ba573f310bf2099e6f9311079e53/R/read_camtrapdp.R#L48-L49

peterdesmet:
Yes please, since self-referencing is simpler now.

peterdesmet (approved):
Thanks! This will be very useful for tests.

I have updated `example_package()` to just default to reading from v1 when an unsupported version is supplied, rather than giving an error. This is a little bit less informative to users, but I plan to have `print()` return the version as found in `datapackage.json` + we don't have to update `supported_versions` in different places.

PR #273, opened 2024-12-28
Allow to add new resource with `replace=true`

### LINKED ISSUE #259: replace=TRUE` prevent the addition of a new resource
Hi,
Is this desired? I would expect that `replace=TRUE` should not prevent the addition of a new resource. 

```r
library(frictionless)

package <- example_package()

# Create a data frame
df <- data.frame(
  multimedia_id = c(
    "aed5fa71-3ed4-4284-a6ba-3550d1a4de8d",
    "da81a501-8236-4cbd-aa95-4bc4b10a05df"
  ),
  x = c(718, 748),
  y = c(860, 900)
)

# Add the resource "positions" from the data frame
add_resource(package, "positions", data = df)
#> A Data Package with 4 resources:
#> • deployments
#> • observations
#> • media
#> • positions
#> Use `unclass()` to print the Data Package as a list.

add_resource(package, "positions", data = df, replace = TRUE)
#> A Data Package with 3 resources:
#> • deployments
#> • observations
#> • media
#> Use `unclass()` to print the Data Package as a list.
```

I didn't look at the function, so not sure why this is happening.

Rafnuss:
Solution for #259.

We could rename all `replace` by `overwrite` but maybe not?


PietrH:
Instead of setting `replace` to `FALSE` when a new resource is added, I made a small change in the way the resource is added to take the replace switch entirely out of the equation. Now it's only used to return an error early, and if it passes that check:

```r
 Check resource does not exist yet for replace = FALSE
  if (!replace && resource_name %in% resource_names(package)) {
    cli::cli_abort(
      c(
        "{.arg package} already contains a resource named
        {.val {resource_name}}.",
        "i" = "Use {.arg replace = TRUE} to replace an existing resource."
      ),
      class = "frictionless_error_resource_already_exists"
    )
  }
  ```
  
  It'll allow for both overwriting or replacing. 
  
  This seems like a more elegant solution to me and fixes the bug. The proposed solution seems more difficult to maintain to me because the `replace` switch differs from the user provided value even when no replacement is taking place.

**Review by peterdesmet — APPROVED** [review]
Thanks @Rafnuss, sorry it 18 months to get to this. 😓

Thanks @PietrH, I used your simplified approach, but replaced `purrr::detect_index()` with:

```r
index <- which(resource_names(package) == resource_name)
```

Which I find easier to read and makes use of `resource_names()` which is created for name checking.

PR #274, opened 2024-12-28
Attach resource's data

Rafnuss:
Here is a small suggest to force the data to be attached to the package object rather than stay as path/url. I would find such a functionality quite useful to be able to work on the data directly, rather than having to read it from elsewhere several time.

This could also be implemented as a seperate function if you think it's better.

Or, is there another way to do that?

If you want to implement this, I'm not sure about `col_select = NULL`. Maybe it should be a different value?

Here is may implementation: https://github.com/Rafnuss/GeoLocatoR/blob/master/R/read_gldp.R#L25-L35. As you can see I depend on `frictionless:::get_resource` and `frictionless:::read_from_path` which is not allowed I think. Any alternative would be great!

### DIFF
--- a/R/read_package.R
+++ b/R/read_package.R
@@ -8,6 +8,8 @@
 #' Data Package standard.
 #'
 #' @param file Path or URL to a `datapackage.json` file.
+#' @param attach Attach the resources' data to the package object rather than keeping the path. See
+#' [data location](https://specs.frictionlessdata.io/data-resource/#data-location).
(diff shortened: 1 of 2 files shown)


peterdesmet:
@Rafnuss I think this is a good idea, but would use a more versatile approach for the `attach` parameter in `read_package()`. I've recorded it here: #372

Closing this as stale, since the internal code has deviated quite a lot since. 😄

PR #284, opened 2025-10-07
Use spatial: null not image: null

peterdesmet:
`image` is a Data Package property that is expected to be a string. It's doesn't return a validation error in frictionless-py, but does in [dpkit](https://dpkit.dev/). It's best we use a custom property/ I choose `spatial`.


PietrH (approved):
👍

PR #295, opened 2026-05-28
Update installation instructions + meta

### LINKED ISSUE #290: Update README cf. usethis
The README.Rmd largely makes use of `usethis::use_readme_rmd()` conventions, but deviates from it to. It would be good to have this standardized, cf. https://github.com/inbo/etn/pull/483.

Changes include:

- Use lowercase frictionless even at start of sentence
- Refer to github.com
- Use pak to install
- Use ``` r for installation code

peterdesmet:
Fix #290

The COC is one for all of frictionless, but the file we refer to refers to that one.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -37,20 +37,18 @@ To get started, see:
 
 ## Installation
 
-Install the latest released version from CRAN:
+Install the released version of frictionless from CRAN:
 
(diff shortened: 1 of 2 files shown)


PietrH (approved):

PietrH:
I think you can install from r-universe using pak as well.

PR #297, opened 2026-06-11
Arrays

### LINKED ISSUE #276: Handling of single element arrays in datapackage.json
There is a discrepancy in how single-element arrays are written and read back into the datapackage.json file. Specifically, single-element arrays are being converted into strings when the package is read and then rewritten.

Example:

1. Create a DP:

```
library(frictionless)

temp <- here::here("frictionless")

resource <- data.frame(1:10, 11:20)
dp_v1 <-
  frictionless::create_package() %>% 
  add_resource(resource_name = "numbers", data = resource) %>% 
  frictionless::create_package()

dp_original$project <- list(
  single_element = list(c("media")),
  double_element = list(c("media", "event"))
)

frictionless::write_package(dp_original, file.path(temp, "1"))
```

2. The resulting datapackage.json is as follows (as expected):

```
{
  "resources": [
    {
      "name": "numbers",
      "path": "numbers.csv",
      "profile": "tabular-data-resource",
      "format": "csv",
      "mediatype": "text/csv",
      "encoding": "utf-8",
      "schema": {
        "fields": [
          {
            "name": "X1.10",
            "type": "integer"
          },
          {
            "name": "X11.20",
            "type": "integer"
          }
        ]
      }
    }
  ],
  "project": {
    "single_element": [
      "media"
    ],
    "double_element": [
      ["media", "event"]
    ]
  }
}
```

Read datapackage.json and write it back:

```
dp_v2 <- frictionless::read_package(file.path(temp, "1", "datapackage.json"))

frictionless::write_package(dp_read1, file.path(temp, "2"))
```

The resulting datapackage.json after being rewritten is not as expected:

```
{
  "resources": [
    {
      "name": "numbers",
      "path": "numbers.csv",
      "profile": "tabular-data-resource",
      "format": "csv",
      "mediatype": "text/csv",
      "encoding": "utf-8",
      "schema": {
        "fields": [
          {
            "name": "X1.10",
            "type": "integer"
          },
          {
            "name": "X11.20",
            "type": "integer"
          }
        ]
      }
    }
  ],
  "project": {
    "single_element": "media",
    "double_element": [
      ["media", "event"]
    ]
  }
}

```

sannegovaert:
This PR fixes #276 .


sannegovaert:
The fix is tested with 2 acoustic Camtrap DP: https://github.com/camera-traps/bioacoustics/pull/30

peterdesmet (approved):
Good to have this fixed! I made a number of important changes:

- I've updated the example dataset so that the `missingValues` of the second resource are a single-element array. That way, we can capture the updated behaviour where such elements are retained with the test `write_package() writes unaltered datapackage.json as is`.
- `enum` is now always written as a list. Before, single element `enum` (very rare!) would be written as `"enum": "value"` which is incorrect. Did not create a test, as it's really an edge case (that is now resolved).
- I've described in NEWS that multiple `path` and `missingValues` will now be written on multiple lines. This will affect _us_ when working with the bird-tracking datapackage.json files. 😁

PR #303, opened 2026-07-28
Make resource() function public

### LINKED ISSUE #283: Make `resource()` function public
For developers, it would be useful if the `resource()` function (previously `get_resource()`) was a public function to access a resource (list) by its name.

sannegovaert:
Export and add example, fixing #283 .


peterdesmet (approved):
Nice, I've updated:

- the example to be in line with `schema()`
- the vignettes, so we clearly separate access from read functions
- frictionless is now always written lowercase (something we forgot in the vignettes)

Issue #144, opened 2023-06-26
error in documentation of `write_package`?


hansvancalster:
The documentation of `?write_package()` states in the description section: "Writes a Data Package and its related Data Resources to disk as a datapackage.json and CSV files. Already existing CSV files of the same name will not be overwritten."

But I think it is the other way around: "Already existing CSV files will be overwritten."

Internally, `write_package()` calls `write_resource()` which in turn uses `readr::write_csv(resource$data, file.path(directory, file_name), na = "")`. The latter has `append = FALSE` as default, which occurding to the doc will overwrite existing csv: "If FALSE, will overwrite existing file. If TRUE, will append to existing file. In both cases, if the file does not exist a new file is created."

N.B. I am in favour of keeping the current behaviour, but just want to note that the documentation should probably be fixed. If on the other hand, you actually wanted the behaviour of the documentation, I suggest adding an extra boolean argument `overwrite`.

peterdesmet:
Hmm, the intend is _not_ to overwrite existing data files by default. The common use case is having a number of csv files and wanting to add a `datapackage.json` to the directory to describe those. There you don't want to alter the original data.

1. You are right that the behaviour of `readr::write_csv()` is to overwrite the existing file.
```R
cars <- mtcars
write_csv(cars, "cars.csv")
cars <- cars %>% filter(mpg == 21)
write_csv(cars, "cars.csv")
```

2. In frictionless, oddly, I don't think that is the case. There is a [test](https://github.com/frictionlessdata/frictionless-r/blob/7dba18d98afe46480367ef2f878160efecb7e629/tests/testthat/test-write_package.R#L59-L83) for it. Note that I will remove `observations_1` and `observations_2` from that test, because those are remote files that are never written.

Trying locally, I can replicate files not being overwritten. @hansvancalster have you tested this? Any idea why the behaviour differs from readr?

In any case:

1. We should probably not rely on undocumented/edge behaviour.
2. Any use case for wanting files to be overwritten? If so, we could use an `overwrite=TRUE` (with default `FALSE`)

hansvancalster:
Below is a reprex where `cars.csv` is overwritten (and where I also expect it to be overwritten). This differs however from 

> "The common use case is having a number of csv files and wanting to add a datapackage.json to the directory to describe those."

[...]

PR #304, opened 2026-07-28
Clarify overwrite behaviour (only true for existing path) in `write_package()`

peterdesmet:
Originally, `write_package()` stated:

> Already existing CSV files of the same name will not be overwritten.

This was confusing (#144), as it is only true for CSV files that were there when the package was read (since we don't want to overwriting this that were unchanged by the user). For a newly added resource that is written, updated and written again, the CSV files _are_ changed.

This PR updated the documentation to clarify that "not overwritten" only applies to existing resources with `path`. Note that no changes were made to the functionality (as it is how we want it).

PS: this change does not merit an update in NEWS.md


sannegovaert (approved):
Clear!

PR #314, opened 2026-07-29
Assign resource

### LINKED ISSUE #312: Create `resource()<-` function
Now that we have a public `resource()` function, it make sense to help developers further with a `resource()<-` function, so they can update a resource in place.

We'll need it ourselves for `upgrade_package()` later. It's also useful for tests (#311).

peterdesmet:
Fix #312

Design decisions:

- `resource()<-` does not validate the input, offering flexibility to developers
- `resource()<-` reverts changes made by `resource()` so both functions can be used in tandem


R/resource.R
```diff
@@ -1,12 +1,28 @@
...
+#' For public manipulation of resources, use [add_resource()] and
+#' [remove_resource()].
+#'
+#' `resource()` gets a Data Resource from a Data Package by its name.
+#' The returned value will be a list describing a Data Resource, with a new
+#' attribute `data_location` to indicate how the data are attached.
+#' If present, `path` will be updated to the full path(s).
+#'
+#' `resource<-` overwrites a Data Resource in Data Package with a new value.
+#' The assigned value will typically be obtained with `resource()` and then
```
sannegovaert: The assigne value will typically be a resource that was obtained with `resource()` and then manipulated. Therefore, this assignment function will revert internal changes made by `resource()` (...).

R/resource.R
```diff
@@ -16,11 +32,17 @@
 #' # Get the resource "deployments"
 #' resource <- resource(package, "deployments")
 #' str(resource)
+#'
+#' # Update the resource
+#' resource$title <- "Camera trap deployments"
```
sannegovaert: resource$title already has the title "Camera trap deployments".
Suggestion to change it to something else, eg "deployments"

sannegovaert (approved):
Nice!
I have 2 small suggestions in the comments.

PR #341, opened 2026-08-21
Generic `version()`

### LINKED ISSUE #340: Make `version()` generic for package, resource, schema and dialect
- [x] Report version for Data Package
- [x] Report version for Data Resource
- [x] Report version for Table Schema
- [x] Report version for Table Dialect

peterdesmet:
Fix #340 

No changes to the function were needed, since the version is always detected with the presence and value of `$schema`, which is a property of:
- Data Package
- Data Resource
- Table Dialect
- Table Schema

So this is mainly a change in documentation + an expansion of the tests.

Note that in the last test I don't make use of the convenience functions `resource()` and `schema()` since those functions will automatically upgrade to 2.0 in the future.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -58,4 +58,4 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-Config/roxygen2/version: 8.0.0
+Config/roxygen2/version: 8.1.0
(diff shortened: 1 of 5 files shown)


damianooldoni (approved):
Nice. Especially the fact that you wrote tests also for versions > 2.0 and a release candidate as well.
GitHub noticed a small easy-to-fix conflict with the `main` branch. You need to solve before merging, of course.

PR #343, opened 2026-08-21
Add `update_resource()` helper, no longer require `tabular-data-resource`, use v2 resource properties (`type` and `$schema`)

### LINKED ISSUE #108: Support `resource.$schema` and no longer require `tabular-data-resource`
Changelog: [resource.$schema (new)](https://datapackage.org/overview/changelog/#resourceschema-new)

- [x] Check `$schema`
  - [x] If not provided, look for `profile` (see also #221)
  - [x] If no `profile`, assume v1
  - [x] If v1: investigate what effect is
  - [x] If v2: interpret new properties
  - [x] If v2+: warn that it may not understand all properties -> **Return version as >=2.0
- [x] Update implementation
- [x] Update documentation and mentions of profiles
- [x] Update tests
- [x] Update examples

In addition:

- [x] No longer require `tabular-data-resource` ([removed in this PR](https://github.com/frictionlessdata/datapackage-v2-draft/pull/51)):
  - [x] `read_resource()`
  - [x] `write_package()`
- [x] Update documentation https://docs.ropensci.org/frictionless/reference/read_resource.html#profile
- [x] Update tests
- [x] Update examples

Currently, data resources without the profile `tabular-data-resource` fail:

```
Error: Resource `validation_report` must have property `profile` with value `tabular-data-resource`.
```

I notice that resources without such a profile are quite common in the wild. These generic `data-resource` _can_ use Table Schema in `schema`, but are not required to do so. Maybe `read_resource()` should be a bit more lenient and assess in a different way if a resource can be read. I think it already thus by e.g. requiring `schema` to contain `fields`.

### LINKED ISSUE #221: Support `resource.type`
CHANGELOG: https://datapackage.org/standard/data-resource/#type

- [x] Requires #225
- [x] If not provided, look for `profile` first. If set to `tabular-data-resource`, assume `table`
- [ ] ~If not provided, then detect from file. Or make the frictionless-r decision to always assume `table`~ -> part of #288 
- [x] Implement
- [x] Update documentation (anywhere Tabular Data Resource is mentioned): `schema()`, `read_resource()`, `add_resource()`
- [x] Update tests
- [x] Update examples

### LINKED ISSUE #338: Create `upgrade_resource()` helper function
This helper function should upgrade a v1 data **resource** to v2

- [x] Write in ~`upgrade_resource.R` file~ -> `utils-upgrade.R`
- [x] Figure out when to call this function
  - [x] Ideally not by `read_package()` as it might generate ERRORs for the user that blocks them from fixing it (since they can't read)
  - [x] Maybe each time we call `resource()`? That would mean resources are only upgraded on e.g. `write_package()`, which is probably fine.

peterdesmet:
Fix #338 
Fix #108
Fix #221

**Note for reviewer**: `upgrade(resource = TRUE)` is not described in the NEWS. This will be tackled once we have an entire `upgrade_package()` function.

This PR upgrades resources from v1 to v2. A v2 resource has:

```json
{
  "$schema": "https://datapackage.org/profiles/2.0/dataresource.json",
  "name": "as in v1",
  "path" "as in v1",
  "type": "table" # rather than "profile": "tabular-data-resource"
}
```

## resource.$schema

- Added by new `upgrade_resource()` helper function
- Added by `add_resource()` as the **first property** of a resource (before `name`), to make it immediately clear what version we're dealing with (v1 if undefined).
- Described in data-resource.Rmd
- Added as **first property** to test files that have resources: `mark.json` and `types.json`.

## resource.type

- Added by new `upgrade_resource()` helper function if `"profile": "tabular-data-resource"`
- Added by `add_resource()`, since that function only adds tabular resources
- Described in data-resource.Rmd
- NOT required by `read_resource()` or `schema()`, since the `type` is only MAY in the [spec](https://datapackage.org/standard/data-resource/#type). Those functions still expect a `schema`.
- Added as property to tests files that have resources: `mark.json` and `types.json`.

## resource.profile

- Used by `upgrade_resource()` to set `"type": "table"`. It then removes it.
- No longer set by `add_resource()`
- Described in data-resource.Rmd as deprecated
- NO LONGER required by `read_resource()` or `schema()` (a long-standing [issue](https://github.com/frictionlessdata/frictionless-r/issues/108#issue-1143289035)). Those functions still expect a `schema`.
- All references (in functions and vignettes) to "tabular-data-resource" are updated to v2 spec.

## `upgrade` parameter in `resource()`

`upgrade_resource()` is called by `resource(upgrade = TRUE)`. Initially I thought to always upgrade, but the `upgrade` parameter has three advantages:

- The sibling assignment function `resource()<-` is designed to remove any changes made by `resource()` (such as internal properties) and having `resource()` automatically upgrade would require `resource()<-` to detect that upgrade (impossible) and downgrade. So the default for `upgrade` is `FALSE`.
- In all functions where we use `resource()` we test with and without upgrades.
- In tests, we can use `resource()` with and without upgrades.

## What runs on resource v2 now?

- Any tests that makes use test files `mark.json` and `types.json`.
- Since `resource(upgrade = TRUE)` is not used anywhere, all functions still run on v1 (i.e. those resources are not silently converted to v2). **Since none of the removed properties or added properties are used, all functions handle both v1 and v2.**

## TODO

- [x] Add tests, as part of test-resource.R
- [x] Don't upgrade within functions, let's do all upgrading `upgrade_package()` later
- [x] Update NEWS
- [x] Update `test-version()` to make use of `resource()`
- [x] Assess #108
- [x] Assess #221


damianooldoni (approved):
Thanks @peterdesmet for the clear description of this PR. You applied the changes as described, nice. And I like you moved from `expect_equal()` to `expect_identical()`.

PR #344, opened 2026-08-24
Relax resource_name requirement

### LINKED ISSUE #172: Relax tests for resource names
CHANGELOG: https://datapackage.org/overview/changelog/#resourcename-updated

- [x] Update `add_resource()`:

https://github.com/frictionlessdata/frictionless-r/blob/d659b0b8196a5feb39b58ccc0814a62d434cfe39/R/add_resource.R#L83-L93

- [x] Update tests
- [x] No documentation changes afaik
- [ ] ~Note: this can only be relaxed for v2 data packages, v1 still requires the strict rules~ We don't create v1 anymore

peterdesmet:
Fix #172

### DIFF
--- a/R/add_resource.R
+++ b/R/add_resource.R
@@ -3,8 +3,6 @@
 #' Adds a Data Resource to a Data Package.
 #' The resource will be a [Tabular Data Resource](
 #' https://datapackage.org/standard/data-resource/#tabular).
-#' The resource name can only contain lowercase alphanumeric characters plus
-#' `.`, `-` and `_`.
 #'
 #' See `vignette("data-resource")` (and to a lesser extend
 #' `vignette("table-dialect")`) to learn how this function implements the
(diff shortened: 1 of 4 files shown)


R/add_resource.R
```diff
@@ -111,6 +97,9 @@ add_resource <- function(package, resource_name, data, schema = NULL,
     )
   }
 
+  # Trim leading and trailing spaces in resource name
```
damianooldoni, line 100: I couldn't find it anyhere in datapackage documentation or in #172, but I suppose it's up to me?

damianooldoni (approved):
Nice! Accepted. But I have two questions: 
- trimming white spaces: I didn't find any mention of it. Is it required somewhere? (see comment in code)
- names are unique: I didn't find any test in add-resources about it, but I see that it's a required property of resource names. I think we should test this property when adding a resource. The check itself should maybe happen in check function.

R/add_resource.R
peterdesmet: Trimming is definitely not required. It's just a help for the user.

PR #350, opened 2026-08-25
Migrate `resource()` and `resource()<-` to v2

### LINKED ISSUE #327: Migrate `resource()` and `resource()<-` to Data Package v2
- [ ] Until #318 is implemented, use `frictionless::example_package(version = "2.0")` as input
- [ ] Likely no changes are needed (that would mean this function works for both v1 and v2)
- [x] Run tests
- [x] Add tests:
  - [x] `resource()` supports Data Package v1 and v2
  - [x] `resource()` returns resource in same version as provided
- [ ] If changes were implemented
  - [ ] Update documentation
  - [ ] Update examples if needed
  - [ ] Update NEWS.md
- [x] Update support table in #320

Note: this change might affect other functions, especially https://github.com/frictionlessdata/frictionless-r/issues/323, so that might need to be implemented first.

PietrH:
Fix #327

- I spotted that one test needed the default version of `example_package()´ to be 1.0, so I added the version argument to make this stable 
- I added tests to see if ´resource()´ and ´resource()<-´ return the same resource versions as their inputs.

### DIFF
--- a/tests/testthat/test-add_resource.R
+++ b/tests/testthat/test-add_resource.R
@@ -1,4 +1,4 @@
-test_that("add_resource() returns a valid Data Package", {
+test_that("add_resource() returns a valid package", {
   p <- example_package()
   df <- data.frame("col_1" = c(1, 2), "col_2" = c("a", "b"))
   df_csv <- test_path("data/df.csv")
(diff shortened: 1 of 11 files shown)


peterdesmet (approved):
Nice! I updated the test for assignment function, so that both the returned package and resource are tested for their version.

And this some unrelated cleanup.

PR #354, opened 2026-08-25
Add `schema.missingValues` support in `read_resource()`

### LINKED ISSUE #222: Support labeled `schema.missingValues`
CHANGELOG: [schema.missingValues (updated)](https://datapackage.org/overview/changelog/#schemamissingvalues-updated)

See also #174 

- [ ] ~~Implement with interlacer package and #213~~
- [x] Support labelled missing values
- [x] Add tests
- [ ] Update documentation

Before (still supported!): 

```json
"missingValues": [
          "NA"
        ]
```

But now also:

```json
        "missingValues": [
          {
            "value": "",
            "label": "missing"
          },
          {
            "value": "NA",
            "label": "not applicable"
          },
          {
            "value": "NaN",
            "label": "not a number"
          }
        ]
```

The labels themselves will be lost, the values transformed to `NA`

PietrH:
I've added a test for labelled missingValues. It turns out `read_resource()` already supports this via `read_from_path()`. I'm testing this explicitly on a v2 package, because the example package that I modify already has labelled missing values. It should however support a v1 just the same if it happens to use labelled missing values. 

The added test was heavily inspired by the existing one. 

Fix #222


peterdesmet (changes requested):
I was very surprised `read_resource()` works out of the box. Turns out it doesn't. 😄 Our new test didn't cover it (now corrected).

Missing values is parsed by `read_resource()` here:

https://github.com/frictionlessdata/frictionless-r/blob/bcbd0db78472aa7f342122dbbb12216a255b72ab/R/read_from_path.R#L68

However, if we provide:

``` r
missing_values <- list(
  list(value = "ignore", label = "Can be ignored"),
  list(value = "wrong", label = "1")
)
unlist(missing_values)
#>            value            label            value            label 
#>         "ignore" "Can be ignored"          "wrong"              "1"
```

<sup>Created on 2026-08-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Then the list of missing values contains 4 values, rather than the expected 2. That means that the label `1` is considered NA, updating our deploymentID value `1` to `NA`.

So, we'll need to parse `missingValues` beforehand before we give it to `read_delim()`.

R/read_from_path.R
```diff
@@ -51,11 +51,13 @@ read_from_path <- function(package, resource_name, col_select) {
   }
   skip <- if (dialect$header %||% TRUE) 1 else 0
 
-  # If missingValues has more than 2 levels, assume it is following the standard
-  # and convert to a flat vector.
-  if(purrr::pluck_depth(schema$missingValues) >= 3){
-    schema$missingValues <- purrr::map(schema$missingValues, "value")
+  # Get missingValues (can be labelled values, simple list or NULL)
+  missing_values <- if (purrr::pluck_depth(schema$missingValues) >= 3) {
```
PietrH, line 55: I think it's a good idea to leave a clue as to why we chose magic number 3 here

PietrH: I chose `purrr::pluck_depth()` because it's very safe, but maybe there is a more readable way to put this? 

Do you prefer `"value" %in% names()` ?

peterdesmet: > "value" %in% names()

No, that might target `value` in (invalidly) deeper nested lists.

I think the pluck_depth() is fine. In the comment I mainly want to write what kind of input you can expect according to the spec. I think the if/else is pretty straightforward from there on

peterdesmet (approved):

PR #357, opened 2026-08-25
Upgrade descriptor

### LINKED ISSUE #345: Create `upgrade_descriptor()` helper function
This helper function should upgrade a v1 **package** to v2

- [x] Write in `utils-upgrade.R` file
- [x] Call this function from `upgrade_package()`

peterdesmet:
Fix #345

- Adds tests for `upgrade_resource()`, which should increase our test coverage
- Adds `upgrade_descriptor()` and tests


tests/testthat/test-upgrade_descriptor.R
```diff
@@ -0,0 +1,73 @@
...
+  # Ignore unregistered value
+  p_v1$profile <- "unregistered-package"
+  expect_identical(upgrade_descriptor(p_v1)$`$schema`, v2_profile)
+})
+
+test_that("upgrade_descriptor() updates contributor role to roles", {
+  p_v1 <- example_package(version = "1.0")
+
+  # Defined
+  p_v1$contributors <- list(
```
PietrH, line 60: Here you add a child element called `contributors`, but later on you expect this element to have a parent `$schema`. Correct?

tests/testthat/test-upgrade_descriptor.R
```diff
@@ -0,0 +1,73 @@
...
+    list(title = "First author", role = "author"),
+    list(title = "No role", custom_property = "custom"),
+    list(title = "Jack of all trades", roles = list("a", "b"))
+  )
+  p_upgraded <- upgrade_descriptor(p_v1)
+  expect_identical(p_upgraded$contributors$`$schema`$roles, list("author")) # Added
+  expect_null(p_upgraded$contributors$`$schema`$role) # Removed
+  expect_null(p_upgraded$contributors[[2]]$roles) # Not added
+  expect_identical(p_upgraded$contributors[[3]]$roles, list("a", "b")) # Kept
```
PietrH: By using `$` and `expect_null()` together, this will also pass if `contributors`, `$schema`, `contributors[[2]]` don't exist. I think it's more defensive to use `purrr::chuck()`

PietrH (changes requested):
~~I'm still having a look, one moment please. ~~ Done

tests/testthat/test-upgrade_descriptor.R
PietrH: $schema doesn't belong here, I've removed it

tests/testthat/test-upgrade_descriptor.R
PietrH: I'm avoiding `expect_null()` to check if an element has been removed. As if we are indexing into a parent element that doesn't exist (`$schema`) we'll pass the test.

PietrH: I've made some changes.

tests/testthat/test-upgrade_descriptor.R
```diff
@@ -0,0 +1,82 @@
...
+  expect_in("contributors", names(p_upgraded))
+  ## Has roles been added?
+  expect_identical(
+    purrr::chuck(p_upgraded, "contributors", 1L, "roles"),
+    list("author")
+  )
+  ## Has role been removed?
+  expect_false("role" %in% names(purrr::chuck(p_upgraded, "contributors", 1L)))
+  ## Has roles not been added where there was no role before?
+  expect_false("roles" %in% names(purrr::chuck(p_upgraded, "contributors", 2L)))
```
PietrH: I'm using magic numbers here, we need to keep in mind that we need to change these if we ever change the object we are testing on. 

It seemed like overkill to make this test independent of the order of the nodes of `contributors`

tests/testthat/test-upgrade_descriptor.R
```diff
@@ -0,0 +1,82 @@
...
+
+test_that("upgrade_descriptor() leaves a v2 package as is", {
+  p_v2 <- example_package(version = "2.0")
+  expect_identical(upgrade_descriptor(p_v2), p_v2)
+})
+
+test_that("upgrade_descriptor() removes profile", {
+  p_v1 <- example_package(version = "1.0")
+  p_v1$profile <- "data-package"
+  expect_false("profile" %in% names(upgrade_descriptor(p_v1)))
```
PietrH: Alternative, but I find this harder to read. Possibly more defensive? I'm not sure. 

```suggestion
  expect_false(purrr::pluck_exists(upgrade_descriptor(p_v1), "profile"))
```

tests/testthat/test-upgrade_descriptor.R
```diff
@@ -0,0 +1,82 @@
...
+})
+
+test_that("upgrade_descriptor() sets $schema", {
+  p_v1 <- example_package(version = "1.0")
+  v2_profile <- "https://datapackage.org/profiles/2.0/datapackage.json"
+
+  # Ignore undefined
+  p_v1$profile <- NULL
```
PietrH, line 28: FYI, in `{rlang}` and `{purrr}` paradigm we use `zap()` to clearly communicate to a human reader that we intend to remove an element rather than set it's value to `NULL`. Functionally this makes no difference here.

tests/testthat/test-upgrade_resource.R
```diff
@@ -0,0 +1,70 @@
...
+  expect_identical(upgrade_resource(resource_v2), resource_v2)
+})
+
+test_that("upgrade_resource() removes profile", {
+  resource_v1 <- resource(example_package(version = "1.0"), "deployments")
+  resource_v1$profile <- "data-resource"
+  expect_null(upgrade_resource(resource_v1)$profile)
+})
+
+test_that("upgrade_resource() sets $schema", {
```
PietrH:
```suggestion
test_that("upgrade_resource() can set $schema to default value", {
```

tests/testthat/test-upgrade_resource.R
```diff
@@ -0,0 +1,70 @@
...
+    "https://specs.frictionlessdata.io/schemas/tabular-data-resource.json"
+  expect_identical(upgrade_resource(resource_v1)$type, "table")
+
+  # Don't set type otherwise
+  resource_v1$profile <- NULL
+  expect_null(upgrade_resource(resource_v1)$type)
+  resource_v1$profile <- "data-resource"
+  expect_null(upgrade_resource(resource_v1)$type)
+  resource_v1$profile <- "http://example.com/my-profiles-json-schema.json"
+  expect_null(upgrade_resource(resource_v1)$type)
```
PietrH, line 69: In this case using `expect_null()` to check if `type` has been removed is just as defensive as `expect_false("foo" %in% names(bar))` or `expect_false(purrr::pluck_exists(bar, "foo"))`. 

What do you think is more readable?

PietrH (approved):
I think the original test failures were due to small mistakes in the tests themselves. Are we actually double sure `role` didn't get removed? 

I made some changes to make the tests more defensive. I think they all reflect our intent now. I also made some minor comments towards my opinionated view on unit testing best practices.

tests/testthat/test-upgrade_descriptor.R
PietrH: Feel free to replace if you prefer `purrr::pluck_exists()`

tests/testthat/test-upgrade_descriptor.R
peterdesmet: Sanity check here, why does `package$contributors[[1]]$role` still return a value?

``` r
library(purrr)
package <- list()
package$contributors <- list(
  list(title = "Peter", role = "author")
)

package$contributors[[1]]$role
#> [1] "author"

package$contributors <- purrr::map(
  package$contributors,
  function(contributor) {
    if ("role" %in% names(contributor)) {
      purrr::list_modify(
        contributor,
        roles = as.list(contributor$role), # Make array
        role = purrr::zap()
      )
    } else {
      contributor # Return as if contributor does not have role
    }
  }
)

package$contributors[[1]]$role # Why does this return "author"?
#> [[1]]
#> [1] "author"
names(purrr::chuck(package, "contributors", 1L))
#> [1] "title" "roles"
```

<sup>Created on 2026-08-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PietrH: Because of partial matching! https://search.r-project.org/R/refmans/base/html/Extract.html

peterdesmet: Answer provided by chat:

```
package$contributors[[1]]$role
```

uses partial matching:

> name
A literal character string or a name (possibly backtick quoted). For extraction, this is normally (see under ‘Environments’) partially matched to the names of the object.

Use:

```
package$contributors[[1]][["role"]]
```

or purrr::pluck(), or `options(warnPartialMatchDollar = TRUE)` to avoid.

tests/testthat/test-upgrade_resource.R
peterdesmet: For now, I'll keep `expect_null()`. We can move to more defensive `purrr` approach later.

tests/testthat/test-upgrade_resource.R
peterdesmet: Good idea to be explicit, will update.

PR #362, opened 2026-08-26
Add S3 `append()` method Data Package objects, ensure `$schema` ends up in first position for `upgrade_` functions

### LINKED ISSUE #198: `append()` drops custom `datapackage` class
When using `append()`, the datapackage lists looses its custom class:

``` r
library(frictionless)
p <- create_package()
class(p)
#> [1] "datapackage" "list"
p <- append(p, c(foo = "bar"))
class(p)
#> [1] "list"
```

<sup>Created on 2024-03-27 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

This results in `check_package()` warnings like in #194. I assume the way to solve this is by creating a `append.datapackage()` function.

To update:

- [ ] vignette, where `append()` is mentioned
- [ ] https://inbo.github.io/movepub/articles/movepub.html

### Overwrite `append()`

Just overwrite `base::append()`. It extends the base functionality, so shouldn't break anything, unless as a user you rely on removing class and attribute properties.

- Good: easy
- Good: will work out of the box on package, resource and schema
- Bad: bad practice
- Bad: start-up warning: "The following object is masked from ‘package:base’: append"

```R
p <- example_package()
p <- append(p, list(title = "My package"), after = 1)
r <- resource(p, "deployments")
r <- append(r, list(description = "Some cool deployments"), after = 0)
resource(p, "deployments") <- r
```

### Extend `append()`

Add a `append.datapackage()` function.

- Good: Good practice
- Bad: only works on package objects. For it to work on resources and schemas (which we want), we would have to add custom classes to those.
- Bad: start-up warning: "The following object is masked from ‘package:base’: append"

```R
p <- example_package()
p <- append(p, list(title = "My package"), after = 1)
r <- resource(p, "deployments")
r <- append(r, list(description = "Some cool deployments"), after = 0) # REMOVES ATTRIBUTES
resource(p, "deployments") <- r
```

### Call it something else

Rather than calling the function `append()`, we could call it e.g. `insert()`

- Good: no start-up warning
- Good: easy
- Good: dplyr tends to do this: `arrange()` rather than `sort()`, `distinct()` rather than `unique()`
- Bad: less discoverable for users

```R
p <- example_package()
p <- insert(p, list(title = "My package"), after = 1)
r <- resource(p, "deployments")
r <- insert(r, list(description = "Some cool deployments"), after = 0)
resource(p, "deployments") <- r
```

### Use a different type of function

Rather than passing a `value` to the function, pass a `property name` + `value`.

- Good: no start-up warning
- Good: don't necessarily have to stick to `append()` parameters/behaviour. E.g. it could error if you're overwriting a property or allow to silence that with `replace = TRUE`
- Bad: less familiar to users

```R
p <- example_package()
p <- add_property(p, "title", "My package", after = 1)
r <- resource(p, "deployments")
r <- add_property(r, "description", "Some cool deployments", after = 0)
resource(p, "deployments") <- r
```

### LINKED ISSUE #358: Set `$schema` as first attribute
`$schema` should be the first property of package, resource, schema. That makes the JSON file easier to read.

- When **creating** an object from scratch, that is easy. Just define it first. `create_package()`, `create_schema` and `add_resource()` already do this.
- When **manipulating** an object that already has properties, it's harder (i.e. what the `upgrade_` functions do). You need to use `append()` or alike.

The issue is that `append()` drops any custom attributes and class, which are required for functions that use that object.

``` r
library(frictionless)
p <- example_package()
attributes(p)
#> $names
#> [1] "name"      "id"        "licenses"  "version"   "created"   "spatial"  
#> [7] "temporal"  "resources"
#> 
#> $directory
#> [1] "/Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library/frictionless/extdata/v1"
#> 
#> $class
#> [1] "datapackage" "list"

p <- append(p, list("$schema" = "just added"), after = 0)
attributes(p)
#> $names
#> [1] "$schema"   "name"      "id"        "licenses"  "version"   "created"  
#> [7] "spatial"   "temporal"  "resources"
```

<sup>Created on 2026-08-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

## TODO

- [ ] Is there a way to solve this in `purrr()`? Note that `prepend()` (deprecated) has the same issue.
- [ ] Is there a way to solve this without a helper function?
- [x] Should `frictionless` have its own `append()` function? See discussion at #198
- [x] Once implemented, revert https://github.com/frictionlessdata/frictionless-r/commit/036bb5243b9bea703f01b7a0c61995861add29b5

PietrH:
Fixes #198

I ended up implementing a generic S3 method for `append()` which allowed me to set a S3 method for `append.datapackage()`. However, since `resource` objects don't have a custom class, I had to implement a helper to append lists while retaining attributes. Some attributes can never be kept while appending, such as names (because they depend on the length of the object staying the same, which is not the case for `append()`). Things like dimnames would also break this custom append helper, as I only added an exception for names.

Converting `append()` to a generic S3 method will result in a message when loading frictionless:

```
Attaching package: ‘frictionless’

The following object is masked from ‘package:base’:

    append
```

This can be avoided by not using S3 dispatch at all, and using the internal helper for packages as well. But then users will not be able to call `append()` on a datapackage without dropping attributes. 

Furthermore, I re-added tests for the location of the `$schema` node in `upgrade_resource()` and `upgrade_descriptor()`. Both use the same append helper code, but `upgrade_descriptor()` uses the generic, anywhere in the package you can call `append()` on a datapackage object, if you want to append a list without dropping the object class or attributes, you need to remember to use `append_with_attrs()`. This also has the advantage that users can now call `append()` on a datapackage object, and it'll just work. The same is not true for resources. 

I started work on adding a custom class for resources, but will leave that exercise for the future.


peterdesmet (commented):
This PR:

1. Sets `$schema` as first property, using a new helper function `append_with_attrs()`, with the goal of solving #358.

2. Adds a public `append()` function (using the new helper), with the goal of solving #198.

Those are very different functionalities and I find this to complex to review in one PR.

## Solution

A stacked PR would have worked here, but since we can't rewrite that, I did the following:

1. From `v2`, create new branch `dollar-schema-first` and checkout the relevant files from this PR. Commit with @PietrH as co-author created PR #366

2. From `v2`, create new branch `append-function` and checkout the relevant files from this PR. Commit with @PietrH as co-author created PR #365

I will review those separately. The helper first, since that one is needed for 2.0.0. The second one can wait.

PR #363, opened 2026-08-26
Create `upgrade_schema()`

### LINKED ISSUE #339: Create `upgrade_schema()` helper function
This helper function should upgrade a v1 table **schema** to v2

- [ ] Write in `utils-upgrade.R` file
- [ ] Call this function from `upgrade_package()`

sannegovaert:
- fixes #339 

`foreignKeys.reference[x].resource` is removed when equal to the resource name. I first tried to add it only when it is not equal to resource name, but then the order of the list was changed. Therefore this solution


peterdesmet (approved):
Nice! I've changed to foreignKeys approach to make use of `purrr::modify_if()`, which doesn't remove custom properties (those are used by DwC-DP I believe) and made the tests stricter.

I've also used the `modify_if()` approach for contributors in `upgrade_descriptor()`

peterdesmet (commented):
Oh, and I added a helper function `character_to_list()` to convert string values to arrays, since we have to do that in 4 places.

Issue #167, opened 2024-01-08
Support `groupChar` for integers


peterdesmet:
CHANGELOG: https://datapackage.org/overview/changelog/#integer-field-type-updated

`groupChar` can now be used for **integers** in addition to numbers.

- [ ] Update `read_resource()` (which only supports `groupChar` for numbers):

https://github.com/frictionlessdata/frictionless-r/blob/848ef0263b0e5588e429e7a48a1e59ca2286158f/R/read_resource.R#L306

to (cf numbers):

https://github.com/frictionlessdata/frictionless-r/blob/848ef0263b0e5588e429e7a48a1e59ca2286158f/R/read_resource.R#L297-L299

- [ ] Update tests
- [ ] Update documentation

PR #364, opened 2026-08-26
Group char support integers

damianooldoni:
This pull request improves the handling of `groupChar` and `decimalChar` properties when parsing number and integer fields, ensuring that grouping and decimal marks are correctly validated and applied. It adds stricter validation to prevent conflicting marks, updates the parsing logic for integer fields, and expands test coverage to handle new and edge cases.

**Parsing logic and validation improvements:**
* Updated the `col_integer` function to accept and use the `group_char` parameter, allowing integer fields to support grouping marks similarly to number fields as required by #167 
* Added validation in the `locale` function to raise errors if the grouping mark is the same as the decimal mark, with specific error messages for default and non-default decimal marks.

**Test data and coverage:**
* Added new test resources and corresponding CSV files to cover cases where integer fields have grouping marks and where grouping and decimal marks conflict.
* Expanded tests in `test-read_resource.R` to check parsing behavior and error handling for the new edge cases, including warnings and errors for invalid mark combinations.
* One regexp based test was testing exactly the same as the previous cli class based test: removed (18d23a2)
* Two regexp based tests (the second one was mine, copy pasting the first one already present) were not testing the vroom warning for parsing failure as actually expected in the comment, but testing the again the cli class based warning. So, I slightly updated the comment to make it more clear and updated the regexp expression to check exactly the vroom warning. See a65aa5c.


peterdesmet (approved):
Thanks! I made a number of changes

- No extra info in error for decimal_mark = grouping_mark, since the warnings are still shown to the user, which is sufficient
- No diff in error between default and non-default decimal mark
- Renamed some test files and classes (including existing)

Issue #179, opened 2024-03-13
Support new field type `list` 


peterdesmet:
The Frictionless spec was updated with a new field type "list". See https://github.com/frictionlessdata/datapackage/pull/38 for details. The frictionless R package could not support this (i.e. default to string) or convert values on `delimiter` to a vector.

khusmann:
I'm interested in this field type for representing multiselect items (although it will have to wait until the `list` field type can be extended by the `categories` property)

In the meantime, I'd vote for the latter approach (convert the cell to a vector using `delimiter` and load as list-columns).

For example, the csv:

```
row_id, field1
1, "a,b,c"
2, "d,e"
3, "f"
```

with schema fields:

```json
[
  {
    "name": "row_id",
    "type": "number"
  },
  {
    "name": "field1",
    "type": "list",
    "delimiter": ",",
    "itemType": "string"
  }
]
```

would become:

``` r
library(tidyverse)
tibble(
  row_id = 1:3,
  field1 = list(
    c("a", "b", "c"),
    c("d", "e"),
    c("f")
  )
)
#> # A tibble: 3 × 2
#>   row_id field1   
#>    <int> <list>   
#> 1      1 <chr [3]>
#> 2      2 <chr [2]>
#> 3      3 <chr [1]>
```

<sup>Created on 2024-06-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Issue #170, opened 2024-01-26
Forbid paths starting with `.`


peterdesmet:
The frictionless spec (https://github.com/frictionlessdata/datapackage/pull/19) might forbid paths starting with `.` (in addition to absolute and relative parent paths). If adopted, `check_path()` should be extended:

https://github.com/frictionlessdata/frictionless-r/blob/d659b0b8196a5feb39b58ccc0814a62d434cfe39/R/check_path.R#L17-L19

PR #368, opened 2026-08-27
Support `type = list` (interpret as character)

peterdesmet:
Easiest approach to #179. It supports users who have this type in their dataset rather than throwing an error for unrecognized type.

Leaving #170 for later full support.


PietrH (approved):
Should this be documented in NEWS? In one of the articles? 

Let me know if you need help with the character to list parsing.

PR #370, opened 2026-08-28
Style: do not use equals for global assignment

PietrH:
Small style fix

### DIFF
--- a/R/version.R
+++ b/R/version.R
@@ -39,9 +39,9 @@ version <- function(x) {
   }
 
   # Extract version from e.g.
-  # "https://datapackage.org/profiles/1.0/datapackage.json" or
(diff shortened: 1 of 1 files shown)


R/version.R
```diff
@@ -39,9 +39,9 @@ version <- function(x) {
   }
 
   # Extract version from e.g.
-  # "https://datapackage.org/profiles/1.0/datapackage.json" or
+  # "https://datapackage.org/profiles/1.0/datapackhttps://github.com/frictionlessdata/frictionless-r/pull/366age.json" or
```
peterdesmet: I assume this is a typo?

peterdesmet (approved):
Fixed an unintentional typo

R/version.R
PietrH: Indeed, oops!

PR #382, opened 2026-09-16
Create `add_properties()` function

### LINKED ISSUE #48: Create function `add_metadata()` or `add_property()`
Add function to add common properties to `descriptor` file, e.g.:

```R
package <- add_metadata(id = "https://doi.org/10.5281/zenodo.5070086")
```

- [ ] Would the function check if the parameter names are valid? If so, what about optional ones? Maybe `add_property()` is better
- [ ] How would the function react if a property is already there? Overwrite with warning?
- [ ] Can we use the `add_property()` function for adding properties to fields or schemas? If so, can the user give a vector of equal length to the number of properties?

### LINKED ISSUE #198: `append()` drops custom `datapackage` class

peterdesmet:
This function is the result of a discussion on how to best support an `append()`-like function. This implementation has several advantages:

- Does not mask base `append()`
- Works on all frictionless list objects (package, resource, dialect, schema), not just package
- Can deviate from `append()` functionality and arguments. E.g. the to-be-added elements are not provided in a single argument `values` but in `...`

## Implementation details

- The function returns an error if arguments provided in `...` are not named. This is done with helper function `check_dots()` (a renamed `get_dot_names()`) which is also used by `add_resource()`
- The function replaces the internal helper `append_with_attributes()`. Many of its tests were kept.
- The function is mentioned in all vignettes. Phrasing that frictionless does not support list manipulation is removed and the `append()` example in frictionless.Rmd replaced.

## Relation to issues

- Fix #48
- Fix #198
- Close #365 (this function replaces that approach)


R/add_properties.R
```diff
@@ -0,0 +1,63 @@
+#' Add properties
+#'
+#' Adds or inserts provided arguments as elements to a list.
+#' Attributes and classes of the original list are retained.
+#'
+#' @details
+#' `add_properties()` can be used to add (custom) metadata properties to a Data
+#' Package, Data Resource, Table Dialect or Table Schema.
```
sannegovaert, line 8: Can you also add a test on Table Dialect?

R/add_properties.R
```diff
@@ -0,0 +1,63 @@
...
+#'
+#' `add_properties()` has some advantages over:
+#'
+#' - [append()]: removes attributes and classes, invalidating a Data Package
+#'   object.
+#' - Direct assignment (`package$title <- "My package"`): no control over the
+#'   position of the added element.
+#'
+#' `add_resource()` also supports adding metadata properties to a Data Resource,
+#' and checks for reserved properties.
```
sannegovaert: What do you mean with reserved?

R/add_resource.R
```diff
@@ -146,25 +146,20 @@ add_resource <- function(package, resource_name, data, schema = NULL,
...
-  if (length(conflicting_properties) != 0) {
+  conflicting_names <- property_names[property_names %in% reserved_names]
+  if (length(conflicting_names) != 0) {
     cli::cli_abort(
       c(
-        "{.arg {conflicting_properties}} must be removed as argument{?s}.",
-        "i" = "{.field {conflicting_properties}} {?is/are} automatically added
-               as resource propert{?y/ies}."
+        "{.arg {conflicting_names}} must be removed as argument{?s}.",
+        "i" = "{.field {conflicting_names}} {?is/are} automatically added as
```
sannegovaert, line 161: I don't understand this informative message.

vignettes/table-dialect.Rmd
```diff
@@ -38,7 +38,11 @@ frictionless supports most dialect properties to read [Tabular Data Resources](h
 
 ### Manipulate
 
-frictionless does not support direct manipulation of the dialect. `add_resource()` allows to set one property (`delimiter`) when data are provided as a file, all other properties are assumed to be the default.
+A dialect is a list you can manipulate, but that is typically not necessary.
+
+`add_resource()` set one property (`delimiter`) when data are provided as a CSV file with non-default delimiter, all other properties are assumed to be the default.
```
sannegovaert:
```suggestion
`add_resource()` sets one property (`delimiter`) when data are provided as a CSV file with non-default delimiter, all other properties are assumed to be the default.
```

sannegovaert (changes requested):
I really like `add_properties()`, it is so elegant!

R/add_properties.R
peterdesmet: I could, but it's basically the same as a resource or schema. Do you think it's useful?

R/add_properties.R
peterdesmet: Properties that are automatically set by `add_resource()`. Would the following be clearer?

> `add_resource()` also supports adding metadata properties to a Data Resource, except those automatically set by that function.

R/add_resource.R
peterdesmet: It generates this for example:

``` r
library(frictionless)
example_package() |>
  add_resource("iris", iris, type = "table", format = "csv", title = "Iris")
#> Error in `add_resource()`:
#> ! `type` and `format` must be removed as arguments.
#> ℹ type and format are automatically added as resource properties.
```

<sup>Created on 2026-09-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Is this unclear?

R/add_properties.R
sannegovaert: Yes!

R/add_resource.R
sannegovaert: ok, I understand, all good!

R/add_properties.R
peterdesmet: Updated, also in vignette: a1778944075e61616219e9972dacc28d39bba995

sannegovaert (approved):

